PR #261, opened 2014-09-04
Updated Pushbullet to use PB API v2, and added PB as a URL sharing service

BallisticLingonberries:
Hey there Jaex, hope you're doing well :D.

I've learned a bunch more about SourceTree and git in general since we last spoke (and I used what you showed me to add support for the API v2).

It took me a while, but I finally got it done (and made some improvements along the way).

Hope you like them :-).

~Karl


micsen:
Hey, 

I would love to see a feature to send the images to a contact instead of only my devices so i could set up for example projector_E330@myschool.com instead of spreading my access key around to everyone who need it. I am sure it wouldn't take much implementing from your side to add the contacts to the list of devices as well. Thanks in advance :)

Issue #372, opened 2014-11-14
FTP Upload destinations don't support pre-escaped user input & non-ascii characters


SolidX:
The current name parser strips out % signs preventing escaped input from being used.

![2014-11-13_22-38-36](https://cloud.githubusercontent.com/assets/1513109/5040973/dabff3c2-6b85-11e4-8f77-331b4f367c7e.png)

lggomez:
A workaround could be making the parserType overridable (in FTPAccount.cs): 
```csharp
        public string GetSubFolderPath(string filename = null, NameParserType nameParserType = NameParserType.URL)
        {
            string path = NameParser.Parse(nameParserType , SubFolderPath.Replace("%host", Host));
            return URLHelpers.CombineURL(path, filename);
        }
```

Then, specify the representation as needed. ex: GetFtpPath()

```csharp
return URLHelpers.CombineURL(FTPAddress, GetSubFolderPath(filemame, NameParserType.FilePath));
```

This may not be an elegant solution but it seems promising, need to check if any additional steps are needed for the case of filepaths. And leaving NameParserType.URL as default seems to be the best choice (since it was the one being used until now)

Jaex:
@SolidX can you test is Iggomez merge fixed your issue?

Jaex:
Close this then.

PR #431, opened 2015-01-01
Fix issue #372

lggomez:
Several changes were made to fix the encoding of ftp dirpaths and allow
uploads to folders with non-ascii characters:

- [x] Fixed encoding of valid urls in ShareX.HelpersLib.NameParser.cs
- [x] Fixed subfolder path generation and wrong url re-encoding in
FTPAccount.cs
- [x] Fixed preview paths in FTP Destination settings (FTPAccount.cs)
- [x] Fixed subfolder path generation in FTP.cs

This fixes issue #372, which was reproduced with the following account
config on a local ftp server (import from clipboard on destination
settings):

```json
{
"Protocol": "FTP",
"Name": "New account",
"Host": "localhost",
"Port": 21,
"Username": "test",
"Password": "test",
"ServerProtocol": "ftp",
"SubFolderPath": "DEV/grhh本人%",
"BrowserProtocol": "http",
"HttpHomePath": "",
"HttpHomePathAutoAddSubFolderPath": true,
"HttpHomePathNoExtension": false,
"IsActive": false,
"FTPSEncryption": "Explicit",
"FTPSCertificateLocation": "",
"Passphrase": null,
"Keypath": null
}
```

### DIFF
--- a/ShareX.HelpersLib/Helpers/Helpers.cs
+++ b/ShareX.HelpersLib/Helpers/Helpers.cs
@@ -43,6 +43,7 @@ using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Windows.Forms;
+using System.Web;
 
 namespace ShareX.HelpersLib
 {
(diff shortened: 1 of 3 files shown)


Jaex:
Thanks. I hope it not broken anything else while fixing encoding.

PR #435, opened 2015-01-04
Added URL decoding to fix #249

### LINKED ISSUE #249: A problem with uploading from clipboard G+ images URL
Hello,
Google plus have long image URLS (too long for ShareX, I guess) so when I am trying to use clipboard upload I have this exception message: http://j.mp/VKc9ep

**English translation (Google Translate):** http://j.mp/VKcfCE

**URL I am trying:** https://lh5.googleusercontent.com/-ecZVg1pjCxI/U_mJ4SO-62I/AAAAAAAA0ZQ/9I8RoCx3n38/w426-h320/%25D0%259F%25D0%25BE%25D1%2587%25D0%25B5%25D1%2581%25D0%25B0%25D1%2582%25D1%258C%25D1%2581%25D1%258F%2B%25D0%25BC%25D0%25B0%25D0%25BB%25D0%25BE%25D1%2581%25D1%2582%25D1%258C%2B%25D0%25BD%25D0%25B0%25D0%25B4%25D0%25BE.gif (I got this url in my G+ Feed)

kopimi:
Added URL decoding before file upload to avoid
System.IO.PathTooLongException when using the URL uploader with an
encoded string

### DIFF
--- a/ShareX/UploadManager.cs
+++ b/ShareX/UploadManager.cs
@@ -454,6 +454,22 @@ namespace ShareX
                         {
                             using (WebClient wc = new WebClient())
                             {
+                                string oldURL, oldPath;
+
+                                do
+                                {
+                                    oldURL = url;
+                                    url = HttpUtility.UrlDecode(url);
+                                }
+                                while(url != oldURL);
+
+                                do
+                                {
+                                    oldPath = downloadPath;
+                                    downloadPath = HttpUtility.UrlDecode(downloadPath);
+                                }
+                                while (downloadPath != oldPath);
+
                                 wc.Proxy = ProxyInfo.Current.GetWebProxy();
                                 wc.DownloadFile(url, downloadPath);
                             }


Jaex:
Your code can stuck in loop forever.

kopimi:
Sorry, under what use case could an infinite loop occur? Just so I don't make the same mistake again. Thanks

Jaex:
Also doing loop like that can be problem like if file name is %50.jpg then it will decode it too, isn't it?

kopimi:
As far as I can tell it's impossible to support multi-encoded URLs while simultaneously protecting against pseudo-encoded file names. Either you loop to support multiple encoding and break files like %50.jpg or you give up the ability to decode URLs that have been encoded multiple times. By definition %50.jpg is an encoded URL which makes it virtually impossible to detect as being a decoded string. I think Unicode filenames like the G+ example and multiple encoded strings which are widespread on many hosts are a more popular use case than %50.jpg style names.

PR #437, opened 2015-01-04
French Translation

nwies:
First French Translation.
I manage to translate everything.


Jaex:
So you don't wanna talk in messenger?

PR #451, opened 2015-01-11
Simplified Chinese（zh-cn） upload

jiajiechan:
Simplified Chinese（zh-cn） upload

### DIFF
--- a/ShareX.HelpersLib/Colors/ColorPickerForm.zh-CN.resx
+++ b/ShareX.HelpersLib/Colors/ColorPickerForm.zh-CN.resx
@@ -1,0 +1,170 @@
+<root>
+  <!-- 
+    Microsoft ResX Schema 
+    
+    Version 2.0
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
+    associated with the data types.
+    
+    Example:
+    
+    ... ado.net/XML headers & schema ...
+    <resheader name="resmimetype">text/microsoft-resx</resheader>
+    <resheader name="version">2.0</resheader>
+    <resheader name="reader">System.Resources.ResXResourceReader, System.Windows.Forms, ...</resheader>
+    <resheader name="writer">System.Resources.ResXResourceWriter, System.Windows.Forms, ...</resheader>
+    <data name="Name1"><value>this is my long string</value><comment>this is a comment</comment></data>
+    <data name="Color1" type="System.Drawing.Color, System.Drawing">Blue</data>
+    <data name="Bitmap1" mimetype="application/x-microsoft.net.object.binary.base64">
+        <value>[base64 mime encoded serialized .NET Framework object]</value>
+    </data>
+    <data name="Icon1" type="System.Drawing.Icon, System.Drawing" mimetype="application/x-microsoft.net.object.bytearray.base64">
+        <value>[base64 mime encoded string representing a byte array form of the .NET Framework object]</value>
+        <comment>This is a comment</comment>
+    </data>
+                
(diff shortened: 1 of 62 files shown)


Jaex:
Thanks for translation! Can you contact me from Google Talk (Hangouts) like using Gmail web messenger. Because for testings (incase I need to resize few controls etc.) and be able to reach you when I need new translations gonna need communication :)

jiajiechan:
Hi, How do I get to test software to debug language?

Jaex:
https://github.com/ShareX/ShareX/blob/master/ShareX/Enums.cs#L31

and

https://github.com/ShareX/ShareX/blob/master/ShareX/LanguageHelper.cs#L48

Jaex:
So how do I ask questions without messenger.

Jaex:
Hello?

jiajiechan:
Hello, how do you need me?

Jaex:
For ask questions like what I need to write to:



and for check is any control need to be resized to fit your language.

Also everytime before releasing I must contact you for you to update translations.

So, a way to contact is required.

jiajiechan:
简体中文（ZH-CN）

my email m@jjc.im

jiajiechan:
OK, I understand what you say, these two places have been changed.

Jaex:
Would this be okey: 简体中文 (Simplified Chinese)
Because using country code not fits with other languages there because I didn't used code for others.

jiajiechan:
Should be ZH-CN，Because there are many Chinese are also divided into many countries and regions，I saw REXS also use zh-cn。

jiajiechan:
You also possible that

Jaex:
here shows Simplified Chinese too.

jiajiechan:
Is that, Simplified Chinese has also been applied to many countries, such as Singapore, Malaysia, etc., used in each country are different Simplified Chinese, zh-cn Simplified Chinese version refers to the Chinese mainland with.

jiajiechan:
prc refers to the Chinese domestic (excluding Hong Kong and Taiwan)

Jaex:
Problem is many people won't know what is zh-cn means. Am I wrong?

jiajiechan:
You can also use the "Simplified Chinese", not much problem, several simplified Chinese language is only part of the syntax is slightly different.

jiajiechan:
There is no problem using the Simplified Chinese

Jaex:
I wonder how much similar zh-CN to zh-SG. If very similar then wouldn't be better use "zh-CHS" so if someone using Chinese Simplified Singapore in his Windows then ShareX will use that language automatically because zh-CHS is parent of both zh-CN and zh-SG ?

jiajiechan:
yes

Jaex:
Then I'm gonna rename zh-CN files to zh-CHS.

jiajiechan:
In fact, I feel no need to share so meticulous, and looked the other software such as photoshop, windows, basically only into Simplified Chinese (zh-cn), Traditional Chinese (zh-tw), Traditional Chinese (zh-hk), other such as Simplified Chinese (Singapore) no one to translate, because they are elected Simplified Chinese (zh-cn) you can see to understand.

Jaex:
Are you know how to use Visual Studio? If so you can test your language now. Otherwise I can compile and upload setup file for you.

Jaex:
History form not opens because "历史:{}0" it should be {0}
fixed it now. Is there a possibility you used {0} {1} things wrong? Because it will crash ShareX.

jiajiechan:
I will not Visual Studio, or you compile good to me, {0} I will fix this problem later.

Jaex:
I already fixed it and gave link to setup which includes your language, so you can test it.

jiajiechan:
Because not previously used github, now know how to update the upstream.

Issue #356, opened 2014-11-02
hubiC Support


ThePierrezou:
Hello,
Can you add a support for hubiC please.
It's a file hoster with 25 GO free :)
( i don't know but i think it can be help you : https://api.hubic.com/ )
Thanks 
ThePierrezou

Fatih54102:
+1

Jaex:
Added by @Kouga-

R0GGER:
Can not find Hubic in Destinations...
screenshot: http://imgurl.nl/2l

McoreD:
@R0GGER reasoning is explained here https://github.com/ShareX/ShareX/pull/452

PR #452, opened 2015-01-12
Add Hubic.com file uploader support

Kouga-:
requested in #356


Kouga-:
requested in #356

Jaex:
Thanks, good work :)

Jaex:
I wonder why you used WebClient. Because none of the uploaders using WebClient in ShareX and have built in methods for sending request which handles progress reporting and other stuff.

Jaex:
Now I tried uploading to folder and it works fine :) But only have one problem that images don't show thumbnail and have no preview. So I will check it now.

Jaex:
Unfortunately I can't find how to show thumbnail and support image preview like normal sharing.

Also I noticed that I can't set root folder to upload.

Jaex:
I wish you checked OneDrive as example folder listing :D Because it allows to select root folder too:

Jaex:
So are you planning to make "Root folder" selectable some how?

Jaex:
I noticed big problem, OAuthInfo don't contains refresh token and it expires. So people need to re auth everytime because it can't refresh itself. So it is completely unusable currently.

Jaex:
Then I need to disable this uploader before release because I can't release broken uploader.

neico:
@Jaex Makes me wonder why you didn't fix the code up instead of just "disabling" it.

Especially after such a long timeframe where it's still not up there...

Issue #436, opened 2015-01-04
@ character appears to be improperly percent encoded when URL is given to clipboard


irandms401:
Uploading a file through (maybe just through FTP?) appears to cause an issue if the filename has an @ in it somehow.

example: uploading a file named "@.png" gives the following URL:

http://blah blah blah/%2540.png

@ percent encoded is %40, and % percent encoded is %25. 
It seems like the percent sign in %40 is getting encoded.
This looks like it happens every single time @ appears, as so:

Link given to my clipboard after upload: http://irandms.noip.me/~irandms/sharex/2015/01/03_19-01-32_CUsersirandmsDocuments%2540_test__test_%2540_test_.png

Actual link: http://irandms.noip.me/~irandms/sharex/2015/01/03_19-01-32_CUsersirandmsDocuments@_test__test_@_test_.png

EDIT: this appears for way more than just @, it pretty much looks like it affects anything that gets percent encoded.

example: file named "! @ # $ % ^ & ( ) [ ] ; ' . , .txt" is put in clipboard as:

http://irandms.noip.me/~irandms/sharex/2015/01/!%20%2540%20%2523%20%2524%20%20%20%2526%20(%20)%20%5B%20%5D%20%253B%20'%20%20.%20%252C%20.txt

instead of the actual URL:

http://irandms.noip.me/~irandms/sharex/2015/01/!%20@%20%23%20$%20%20%20&%20%28%20%29%20%5B%20%5D%20%3b%20%27%20%20.%20,%20.txt

Apologies if this has been reported before, didn't see it in search. Can give more details if needed as well.

Jaex:
@SolidX is your changes caused this issue too :( ?

SolidX:
@Jaex I'll check it out as soon as I get a chance. :+1:

Jaex:
Thanks :)

SolidX:
@Jaex looks like this is still an issue after #431 was merged in.

PR #453, opened 2015-01-12
Fix issue #436

Kouga-:
Fix issue #436 

This return statement was double encoding filename hence the extra encoding on the "%" sign in the URL in clipboard.

return Uri.EscapeUriString(httpHomeUri.Uri.ToString());


Jaex:
@SolidX was added that code in #371
So pull requests when fixing one thing in the other hand breaking other things :(
So I dunno which pull request is correct.

Kouga-:
Hmm....@SolidX is right, I'll see if there is another way to fix it.

Kouga-:
That last commit should fix both #436 and #371.

Jaex:
Thanks

PR #458, opened 2015-01-15
zh-cn change

jiajiechan:
zh-cn change


Jaex:
I think I should write more english thing there instead "ZH-CN"? Maybe Simplified Chinese?

PR #459, opened 2015-01-15
add zh-cn language

jiajiechan:
add zh-cn language


Jaex:
It was not case SupportedLanguage.French :P But no problem I can fix it. I was just asked what should I use as chinese word for that menu.

PR #465, opened 2015-01-16
Simplified Chinese (zh-cn) update

jiajiechan:
Simplified Chinese (zh-cn) update
-Corrected translation appeared in the space problem
-Fixed some translation does not comply with the actual


Jaex:
Because you not pulled from upstream first it is not automatically merge able anymore :(
Probably because I already fixed {}0 problem.

Jaex:
Nevermind I managed to merge with git commands. Still for future you should pull from upstream like this:
http://stackoverflow.com/a/15943715/264877

PR #467, opened 2015-01-16
Updated Simplified Chinese

jiajiechan:
-Add six local translation

### DIFF
--- a/ShareX.UploadersLib/Forms/UploadersConfigForm.zh-CN.resx
+++ b/ShareX.UploadersLib/Forms/UploadersConfigForm.zh-CN.resx
@@ -1,3 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
   <!-- 
     Microsoft ResX Schema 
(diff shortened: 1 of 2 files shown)


jiajiechan:
hi, Jaex.
-I found a few places in the sharex missing translation items.
-Also troublesome Jaex compiled about the latest sharex test program to test me Simplified Chinese language.

thanks.

Jaex:
Those texts are custom hotkey descriptions. When you can write your own description there how am I supposed to make them translatable? But why it is Turkish in your ShareX instead English I'm not sure.
Didn't understood your second question.

Jaex:
By default descriptions empty which using hotkey description and it works fine with translation. When I change language those descriptions changing too.

jiajiechan:
I want to get sharex test, test the last update in Chinese

Jaex:
resx files can't work without compile so without Visual Studio you can't make it work :(
By the way I figured why that hotkey descriptions stuck in one language and fixing it now. So description when empty it can use current language hotkey info.

Jaex:
Alright fixed hotkey description problem:

So if hotkey description is empty (emtpy by default) then it will show language based text :)

Jaex:
Latest beta: https://dl.dropboxusercontent.com/u/14076298/ShareX/2015/01/ShareX-9.6.0-setup.exe
You must edit descriptions of hotkeys to be empty. It is by default empty now but because you using your config it won't be empty for you.

PR #469, opened 2015-01-17
Create Resources.pt.resx

tiagodj7:
Translation into Portuguese


Jaex:
Hi, can you please pull request all resx files together instead one by one? Because otherwise you need to do around 60 pull request.

PR #476, opened 2015-01-19
Czech Translation

Baegus:
This should be a complete Czech translation of ShareX.
I only translated ShareX itself for now, but I'll get to all the other libraries later.


Jaex:
Seriously one commit per file?!

Issue #404, opened 2014-12-12
Google Drive file-upload not working


greensn0w:
Whenever I try to upload a file to Google Drive using the `send to` menu, ShareX gives me this error:

> Response:
Metadata part is too large.

>StackTrace:
   at System.Net.HttpWebRequest.GetResponse()
   at UploadersLib.Uploader.UploadData(Stream dataStream, String url, String fileName, String fileFormName, Dictionary`2 arguments, NameValueCollection headers, CookieCollection cookies, ResponseType responseType, HttpMethod method)

Screenshots are working normally...

TheKeyLimePie:
Did you check your anti-virus program? Yesterday my Norton program decided to block any ShareX traffic.

fernmndez:
It's to do with the actual upload itself, it'd give a different error code / be unreachable if they're being blocked by the anti-virus

mathemans:
![2015-02-12_08-55-00](https://cloud.githubusercontent.com/assets/1627956/6159750/0782615a-b295-11e4-95e1-ab93250379fa.png)

When I upload .ai files, google drive return errors and cannot upload at all.

PaulBGD:
The original 400 error is occurring with me.

Jaex:
Kesmy pull request fixed these issues I think.

PR #516, opened 2015-02-15
Potential fix for #404

matthewlucic:
The addition of the "uploadType=media" has solved the issue of the error
400 when uploading to Google Drive. This however means that it is not a
multipart upload. My research has lead me to believe that the issue
behind it is that there is no metadata being sent with the request.

My suggestion would be to add the metadata as a new part, and as well
change the query parameter to "uploadType=multipart".


Jaex:
I was tried uploadType=media and it was not solution. There was other problems too and also i need to use multipart for be able to set file info and that multipart type was not working too because it require special multi part type which not supported by ShareX. So this API bullshit. They want us to use their libraries. Making it ourself difficult.

matthewlucic:
Fair enough. Hopefully a solution to this problem will unveil itself at some point, or Google will have a more accessible API.

PR #529, opened 2015-02-27
Update simplified Chinese language

jiajiechan:
Update the simplified Chinese language in 9.7.0


Jaex:
This is not mergeable :(

PR #534, opened 2015-03-03
Fixed #532: Filename text length will now check/uncheck 'Filename filter...

### LINKED ISSUE #532: Search field - key listener
On history page can you add event listener on EnterKey for apply filters?
And on first letter inserted in Filename filter set checkbox checked?
http://i.imgur.com/bkTwydL.jpg

Often I write search term, press Enter and find out that I need to take mouse and klick on checkbox and Apply filters.

ElectricDanDan:
...' checkbox. 'Apply filters' button now default accept button.

Not sure why GitHub is showing it as wanting to change hundreds of lines in the Designer and resx files, when it's actually just a few lines here and there. Perhaps it's that I've got Visual Studio set to use tabs instead of spaces?

### DIFF
--- a/ShareX.HistoryLib/HistoryForm.Designer.cs
+++ b/ShareX.HistoryLib/HistoryForm.Designer.cs
@@ -29,246 +29,248 @@
         /// </summary>
         private void InitializeComponent()
         {
-            System.ComponentModel.ComponentResourceManager resources = new System.ComponentModel.ComponentResourceManager(typeof(HistoryForm));
-            this.dtpFilterFrom = new System.Windows.Forms.DateTimePicker();
-            this.cbDateFilter = new System.Windows.Forms.CheckBox();
-            this.lblFilterFrom = new System.Windows.Forms.Label();
-            this.lblFilterTo = new System.Windows.Forms.Label();
-            this.dtpFilterTo = new System.Windows.Forms.DateTimePicker();
-            this.btnApplyFilters = new System.Windows.Forms.Button();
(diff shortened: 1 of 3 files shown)


Jaex:
Yes it is because of tabs and I wrote here therefore: https://github.com/ShareX/ShareX/wiki/Code-contribution
That I only accept pull requests with spaces.

Also this not solves #532 because he want search to happen when typing but yours just checking it. So it is not what he wants.

ElectricDanDan:
Okay, well I will change Visual Studio to use spaces instead.

If you read #532 again you'll see that this commit does actually meet his request. He doesn't ask for a history search, he just wants the 'Filename filter' checkbox to be checked when there is text in the filename textbox. He just wants to be able to type a filter and press enter.

I will change to spaces and make another pull request of the same solution, which will meet his issue and not affect performance.

Jaex:
"On history page can you add event listener on EnterKey for apply filters?"

Which means when you press key it will apply filters. So this pull request not meet his request.

ElectricDanDan:
"On history page can you add event listener on **EnterKey** for apply filters?"

He wants to be able to press the enter key on his keyboard and have it behave the same as clicking the 'Apply filters' button.

Jaex:
Oh I thought enter key means press key.

MilanRS:
ElectricDanDan is right. I don't know how you name that EnterKey, Return key, that key for new line and form submit. :)

Jaex:
I'm reverting `AcceptButton` change because it not allows list to open URLs with <kbd>Enter</kbd>.

PR #539, opened 2015-03-09
Memory Leak Fixes

yankejustin:
Fixes nearly all memory leaks in the application by implementing IDisposable when needed, correctly implementing IDisposable, disposing objects that should be disposed, and adding try/finally blocks to ensure a resource doesn't hang if an exception is thrown.
Some general code structure improvements were made for clarity.


Jaex:
Wouldn't it be better if you did this pull request to Greenshot project?

yankejustin:
I apologize for including the Greenshot ones... I only meant to include the ShareX changes.
As for the changes in ShareX, they were added to correctly release resources that aren't being disposed.

Jaex:
Example in here you disposing open form:



Also you disposing BackgroundWorker in many places but reason why it have IDisposeable because it inherits Component. Which allows you to add BackgroundWorker to forms. But it actually don't have anything which require dispose. So it should not be disposed.

So I can't see any leaks I forget.

PR #545, opened 2015-03-11
Update Resources.de.resx

Pandafuchs:
Fixed a grammar error. 
"Aufgabe: Keiner" makes no sense but "Aufgabe: Keine" is correct.
"Aufgabe Keiner" means Task: Noone and "Aufgabe: Keine" means "Task: Nothing"


Jaex:
You could add this commit to previous pull request.

PR #577, opened 2015-03-30
Initial commit to Spanish Translation

ovnisoftware:
I fixed a typo in the English ApplicationSettingsForm.resx


McoreD:
It is not a typo. Might just need inverted commas to clarify.  

Use "secondary uploaders order of preference" when retrying 
-- means, when retrying, use secondary uploaders. Otherwise just fail the task.

Use secondary uploaders in order of preference when retrying
-- means, use the order of preference in secondary uploaders. This is what be default happens.

ovnisoftware:
Ok, thanks for clarifying it for me.

PR #586, opened 2015-03-31
#404 - Use multipart/related for GDriveSDK uploads

Kesmy:
Updates GoogleDrive & Uploader to include the metadata in a multipart/related when sending to Google Drive. May also fix some cases where files uploaded to gdrive were broken.


Jaex:
Thanks, using multipart/related fixed many of issues I think.

PR #592, opened 2015-04-01
Added Korean translation of ShareX.

123jimin:
Main ShareX project is fully translated; other ShareX projects like ShareX.HelperLibs are partly translated.


Jaex:
Thanks.

PR #601, opened 2015-04-05
Add support for phpuush

ProxyChain:
Adds support for phpuush as a destination (https://github.com/ProxyChain/phpuush).

Originally intended as an alternative server back-end to the puush client, but works well for ShareX with any files.


ProxyChain:
Just thought I'd post up some examples of what phpuush does (beyond storing files):

Syntax highlighting with GeSHi: 
http://blake.kiwi/IUA1/ab7cfe => http://blake.kiwi/IUA1/ab7cfe/php
http://blake.kiwi/I0eu/52cbbc => http://blake.kiwi/I0eu/52cbbc/cpp

QR encoding:
http://blake.kiwi/HFDr/43e38c => http://blake.kiwi/HFDr/43e38c/qr

These are just the base extensions/handlers that I include with phpuush, but it can be extended pretty easily.

Jaex:
"phpuush works by intercepting requests to the puush service and not forwarding them."

Why would I add such thing to ShareX not sure.

ProxyChain:
I'm not sure if you've quite understood what phpuush is and dismissed it prematurely; don't be fooled by the name, it is not "locked down" to the puush client.

Originally it was written to be compliant with the puush protocol and work seamlessly with the puush client, and it allowed users to "get around" the 200 MB file upload limit that was imposed by the official puush website.

However it is a fully-functional file storage platform built on PHP - I don't feel that it is fair that you've dismissed it with "Why would I add such thing to ShareX not sure".

phpuush is widely used among fans of the puush client that don't want to deal with the official puush servers.

Especially after the recent puush malware that was sent out (http://www.zdnet.com/article/puush-calls-for-password-change-after-malware-hit/) a LOT of puush users are switching to ShareX, including me - this is actually how I heard about the client.

I wrote this so that phpuush users wouldn't be left in the dark if they switched to ShareX, and I imagine many more will make the switch to ShareX once they can use phpuush with it again. It is a good tool for people who want to store their own files on their own server and add extensions to manipulate those files (e.g. QR, syntax highlighting as shown previously).

So please, I ask that you reconsider merging this change.

ProxyChain:
I wanted to further add that, phpuush does not "intercept requests to the puush service" as such - its code is written so that ANY client using the same protocol can work with it, as I've done with this ShareX destination code.

The "intercept" part comes from the fact that the official puush client only has "Proxy" settings, so we use these to redirect traffic through phpuush instead - this IS NOT required for ShareX, and it works absolutely seamlessly using the configuration menu (as other ShareX destinations do).

McoreD:
@ProxyChain after the recent puush malware that was sent out, puush users are not left in the dark. ShareX allows uploading to puush using Custom Uploader interface. Have a look here: https://github.com/ShareX/ShareX/wiki/Custom-Uploader-examples

phpuush on the other hand gives just 137 results in Google search. There is no convincing evidence that this is widely used. ShareX is open source and you are quite welcome to use the base of ShareX and fork it for your personal modifications.

PR #607, opened 2015-04-09
Spanish - Shortened some names on MainForm

ovnisoftware:
Please adjust the menu size to fit the new names.


Jaex:
I adjusted menu.

ovnisoftware:
Please let me know if there's anything else I can do to improve the application.  I think it's a really great tool and I am proud to have worked on it.  As I continue to use it I will see if there are any other translation adjustments that need made.  Thank you so much for your quick replies on everything.

Jaex:
I wonder why you used standalone Resx Manager software instead using Visual Studio extension of it. Because I see you are already using Visual Studio, so it would be more easier for both of us. Because standalone software unable to add resx files to project therefore I needed to add them one by one to project. But extension adding them automatically.

ovnisoftware:
I wasn't aware of the Visual Studio extension for ResX Manager.  Sorry you had to add the files manually.  I guess it's kind of late now, but I will install the extension for ResX Manager to check it out.

PR #614, opened 2015-04-10
Include new RESX files in projects

tom-englert:


Jaex:
Sorry for your wasted time but those were uncomplete translations and I only add them when translation is complete because so many times translations left not finished and after that removing them from projects manually would be problem therefore I was not added them :(

But thanks anyway.

tom-englert:
No problem, it was no extra work but just a side effect of testing if the new features in ResXManager work.

PR #630, opened 2015-04-15
 Task for capturing specific monitor

### LINKED ISSUE #594: Task for capturing specifing monitor
Currently you only have tasks for capturing all monitors / screens or the active one.
But you can't say hey please capture Monitor / Screen X.

danielbrezoi:
Task for capturing specific monitor region  fixed #594


Jaex:
I didn't understood few things in codes, example these properties:

MonitorName, MonitorBounds.

MonitorName supposed to be ToString method. It is not like Monitor can have name.

MonitorBounds, I don't even know about that because you already have x, y, width, height so what is point of this?

MonitorRegion class feels very useless to me when you can just use Rectangle?

Also I don't understand how you can give such control names in big form:




Especially when that form have many width related controls etc.

Also this very fragile DefaultScreenRegionCboIndex because monitor count and resolutions can change. So it is no point saving monitor index at all, because we just using custom region anyway which is not connected to monitor?

danielbrezoi:
Thanks for the review.

Setting the labels names was my five minutes of stupidity/day.
The purpose of the monitor object is to offer the possibility to take the bounds from a selected monitor and to put them in our region bounds, that is the only object that will saved.
The monitor name is used to identify the selected object and to override the ToString() method. 

I've chose to create a MonitorRegion object because:
1. It is easy to take all the monitors and to put them in the combo box.
2. It is easy to take the bounds from the selected monitor and to put them in the custom region rectangle.
3 The MonitorName makes some sense for the user., otherwise the user will see only some Rectangle properties. 
3. We create some of the logic out side the form.
Therese no need to have the x, y, width, height  properties. It can be kept only the Bounds property. ore  the other way around.

So you purpose to eliminate the MonitorRegion object and to work with rectangle?
If yes, what do you think we should add in the MonitorBounds ComboBox? I can add the rectangle from each monitor.

After work I'll make the changes an commit a new version.

danielbrezoi:
Can you please review the changes I've make? If you want I can create a version without the MonitorRegion object.

Jaex:
Thanks all looks good now but have one question "[DefaultValue(0)]" how rectangle can have 0 as default value?

Very small typo in resx file: <value>Monitor{0} (X:{1} Y:{2} Height:{3} Width:{4}</value>
Should be <value>Monitor {0} (X:{1} Y:{2} Height:{3} Width:{4})</value>
Was missing ) end of it.

danielbrezoi:
Thanks for the review, I'll change the default value and the missing ')'.
I'll start the localisation but I wanted to ask witch translation engine to use? It's OK to use the Google translator?

Jaex:
What localisation?

danielbrezoi:
It is the same thing as localization. 
Each form in the project has resource files(.resx) for 9 different languages where the displayed text has the translated version. 
I want to input in the string resource files the translated version for the text that I've added in the forms. (ex. "Select a monitor for a custom region:" in french should be "Sélectionnez un moniteur pour une région personnalisée:" )

Jaex:
It is translators work. Never use Google Translator in anything it makes texts completely not understandable.

danielbrezoi:
I've made the two changes. Thanks for your support.

Jaex:
Merged it thanks.

Issue #561, opened 2015-03-17
New authentication schemes in S3 (V4)


SimWhite:
I'm trying to use new Frankfurt region with AWS uploads and getting following error:

<pre>The authorization mechanism you have provided is not supported. Please use AWS4-HMAC-SHA256.</pre>

As I know AWS4-HMAC-SHA256, also known as "V4," is one of new authentication schemes in S3. US-Standard and practically all other regions (but not all of them) also support the other, older scheme, V2.

According to http://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html new S3 regions deployed after January 2014 will only support V4. Since Frankfurt region was introduced in late 2014 it apparently does not support V2.

Jaex:
@alanedwardes can you check it?

alanedwardes:
@Jaex Yeah this looks like something we need to do. I'll take a look at implementing at some point, hopefully the weekend (unless anyone else wants a go at it).

alanedwardes:
Looking into this. It's actually a little trickier than it seems, we need to have access to the AWS region name to sign requests, right now we just have an arbitrary endpoint: http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-authentication-HTTPPOST.html

So this will require a UI change.

alanedwardes:
Also worth noting, this will kill DreamObjects support, they do not support this form of authentication in their API: http://docs.dreamobjects.net/s3-api/authentication.html

Jaex:
Is this region works #605 now or not?

alanedwardes:
no

Jaex:
Then why he added it, if it were not working.

alanedwardes:
That might be a question for him :wink:

Jaex:
@junderhill ?

junderhill:
I didn't realise it only supported v4. Poor assumption on my part, apologies.

PR #636, opened 2015-04-19
Amazon Auth V4 Support

alanedwardes:
Includes changes to the Amazon S3 support to use the `AWSSDK`, and `PUT` instead of `POST`.

Supports the new [V4 authentication](http://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html) mechanism to enable the use of the new Frankfurt region, and future S3 regions.

Fixes https://github.com/ShareX/ShareX/issues/561.


ShareX/Forms/MainForm.cs
```diff
@@ -46,7 +46,6 @@ public partial class MainForm : HotkeyForm
         private bool forceClose, firstUpdateCheck = true;
         private UploadInfoManager uim;
         private ToolStripDropDownItem tsmiImageFileUploaders, tsmiTrayImageFileUploaders, tsmiTextFileUploaders, tsmiTrayTextFileUploaders;
-        private System.Threading.Timer updateTimer;
```
Jaex: Why you removed auto updating?

alanedwardes: This variable wasn't in use

On 20 Apr 2015 8:59 am, Jaex <notifications@github.com> wrote:

In ShareX/Forms/MainForm.cs<https://github.com/ShareX/ShareX/pull/636#discussion_r28669517>:

> @@ -46,7 +46,6 @@ public partial class MainForm : HotkeyForm
>          private bool forceClose, firstUpdateCheck = true;
>          private UploadInfoManager uim;
>          private ToolStripDropDownItem tsmiImageFileUploaders, tsmiTrayImageFileUploaders, tsmiTextFileUploaders, tsmiTrayTextFileUploaders;
> -        private System.Threading.Timer updateTimer;


Why you removed auto updating?

Jaex: It wasn't in use in Debug but in use in Release.

Jaex:
Is DreamObjects support works with this?

alanedwardes:
No, as I said in the original issue, DreamObjects does not support the AWS V4 authentication scheme.

On 20 Apr 2015 9:04 am, Jaex <notifications@github.com> wrote:

Is DreamObjects support works with this?

ShareX/Forms/MainForm.cs
alanedwardes: OK, will restore later.

McoreD:
Need to inform the guys at DreamHost. Users of http://wiki.dreamhost.com/How_to_Use_ShareX_with_DreamObjects aren't gonna be happy.

Jaex:
I'm not really sure is supporting one region more important than removing DreamObjects support :(

alanedwardes:
I think it would be better to poke them to update their authentication schemes, if they’re going for a like for like match of the S3 API, they need to implement the new stuff.

alanedwardes:
It’s entirely up to you – merge it or don’t.

Bear in mind this futureproofs it, any other regions Amazon releases (and they will release more) will use V4 only, so support will gradually drop off. I’m unsure if they will deprecate the V2 authentication scheme for existing regions though, I haven’t read anything that indicates they will.

alanedwardes:
Have tweeted them, will see what they say: https://twitter.com/AlanEdwardes/status/590113749194924032

When I get home I’ll check to see if it’s possible to drop the AWS SDK back down to auth V2 for arbitrary endpoints. I don’t really want to have the old code hanging around just for DO… we could make a separate file uploader for DreamObjects though, that would make more sense since the APIs are diverging. Leave it with me, will update the PR.

alanedwardes:
Closing this while I work on it.

alanedwardes:
The SDK does allow dropping down to the old auth scheme. Re-opened, includes DreamObjects support again :dancers:

Jaex:
Thanks.

PR #689, opened 2015-05-16
Use HTTPS when interacting with Imgur

mileswd:


Jaex:
Imgur web site upload or Imgur api upload on purpose not giving https:// links because in future they can remove support of it and then all links gonna be broken. If they intended to use it always they would use it in website by default. Of course main reason can be bandwidth but still it not change the fact I can't risk using non official url format.

mileswd:
I understand what you mean, but I think it is incredibly unlikely that Imgur will suddenly remove HTTPS support altogether.

For the record, it should be noted that the ```deletehash``` is potentially sensitive information as it can allow anyone to both edit and delete an image - ideally it should only be sent encrypted. Maybe you should consider sending delete requests via the relevant API endpoint instead of just opening the delete link in the browser.

Jaex:
Then you should tell that to Imgur.

mileswd:
@Jaex I'm sorry, I don't understand what you mean.

I'm just making you aware that there is a safer way of deleting images that could be implemented:

Simply send a ```DELETE``` request to ```https://api.imgur.com/3/image/{deletehash}```

Jaex:
I'm telling if you think it is not secure then you should contact Imgur and tell them to return delete url which starts with https in both web site and API.

And if you really think adding delete request support in multi destination supported software that easy, I don't know what to tell. Because it is beyond complex with FTP destination indexes based on file type etc. So many complications.

mileswd:
@Jaex The Imgur API does not return a delete url, it only returns the ```deletehash``` - it is up to the developer to implement the delete functionality, which should ideally be through the API.

I do not know the structure of ShareX; I am just trying to respectfully suggest ways in which security could be enhanced.

On an unrelated note, thank you for your time in reviewing this PR and also for your fast responses

Jaex:
Web site gives deletion url. I forget about API was not giving deletion url because I was implemented it years ago.

Jaex:
#716 fixing your problem too.

mileswd:
@Jaex Appreciated, thank you. :+1:

Jaex:
Added this option now:

PR #699, opened 2015-05-18
Extract Wikipedia article into README

sides:
Extract Wikipedia article into README so this README can be linked to instead, addressing the article's [issues](https://en.wikipedia.org/wiki/Talk:ShareX). If accepted, the Wikipedia article can be stripped of the sections covered here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,155 @@
-### Website: [getsharex.com](https://getsharex.com)
+### [getsharex.com](https://getsharex.com)
 
 [![](https://getsharex.com/img/ShareX_Screenshot.png)](https://getsharex.com)
+
+## Features
+### Capturing
+ShareX incorporates the following methods to allow screen capture:
+ - Fullscreen: Creates a screenshot of the entire screen area.
+ - Active window: Captures the currently active window.
+ - Active monitor: Captures the monitor area where the mouse cursor currently resides.
+ - Window menu: Has list of active windows so user can select which window to take screenshot of.
+ - Monitor menu: Has a list of monitors so the user can select which monitor to take screenshot from.
+ - Rectangle: Allows user to take screenshot from a single rectangle or multiple rectangular areas drawn by the mouse by dragging it from one corner of a rectangle to the diagonally opposite other corner of the rectangle.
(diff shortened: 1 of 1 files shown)


Jaex:
Main reason was why we didn't used GitHub wiki because then we needed to keep both wiki up to date each time need to add something new, which gonna be problem.

Did you used wiki to markdown converter for this?

sides:
> Did you used wiki to markdown converter for this?
Quick regex and manual read-over.

The Wikipedia article can contain a short summary of ShareX's features that doesn't need to be updated, while this README (or Github wiki) is kept up to date. Generally keeping encyclopedic information and the feature set segregated. I can update the Wikipedia article if that's a go.

Jaex:
@McoreD was handling Wikipedia stuff so need to ask his opinion too about it.

Thanks for markdown conversion.

sides:
You may want to consider https://github.com/sides/ShareX/commit/c3056fb6d0742c90abb8202b71a93b4d3dae2497.

I forgot that and don't want to create another PR.

Jaex:
Too late I did changes now it won't be mergeable anymore.

sides:
Oops, well.

Case sensitive replace:
```
user -> you
User -> You
user’s -> your
```

All there is, won't break anything.

Jaex:
Replaced them.

McoreD:
Go for, thanks.

sides:
Done.

There were some bits removed that weren't transferred to the README, but you should be able to find them by diffing the article's versions on Wikipedia.

I otherwise gave the article a full makeover, feel fully free to revert/edit anything that you don't like. It might be under citation overkill right now, I don't know, not an expert. The citation for the features is also listed as this GitHub page which might be more potential COI, although I'm not sure if it counts when it's open-source (where this page is technically the source for the features).

Jaex:
Thanks it looks good.

By the way I was updating features page now and because my English not very good, everytime I ask McoreD to review my changes before I upload them to wiki/markdown. So incase If you would like to help me about such things, are you using any messenger like Google Talk (Google hangouts) or Steam etc. which I can contact?

sides:
OK, yea, I could help a bit. [Steam](http://steamcommunity.com/id/sides) or IRC is good.

Jaex:
I added you in Steam.

Issue #714, opened 2015-05-31
add new uploader


call5:
hello i use your tool and i like it, it's possible to add these uploader ?
- https://img.bi / https://lut.im  for image screenshot 
- https://videobin.org for screencast
thank you for your good job :+1:

k3d3:
@call5 If you want a ShareX uploader that has support for an encrypted image host, we have a fork at https://github.com/Upload/ShareX.

We're planning on getting it merged into mainline ShareX (unless someone gets to it first) but we want to clean up the code a bit first.

@Jaex on that note, would you accept a PR with Up1 support if I made one?

k3d3:
@call5 Si vous voulez, il y a un application d'Up1 en ShareX que vous pourriez utiliser, a cote de img.bi. C'est beaucoup come img.bi et, dans de nombreux cas, est mieux.

Enjoy!

Scrxtchy:
Lut.im
https://lut.im/jWkcEbOP/mXfN6ltj
```
{
  "Name": "lut.im",
  "RequestType": "POST",
  "RequestURL": "https://lut.im/",
  "FileFormName": "file",
  "Arguments": {
    "format": "json",
    "delete-day": "3",
    "first-view": "0"
  },
  "ResponseType": "Text",
  "RegexList": [
    "\"short\":\"(?<short>\\w+\\/\\w+)\"",
    "\"token\":\"(?<token>\\w+)\"",
    "\"real_short\":\"(?<real_short>\\w+)\""
  ],
  "URL": "https://lut.im/$1,short$",
  "ThumbnailURL": "",
  "DeletionURL": "https://lut.im/d/$3,real_short$/$2,token$"
}
```
>delete-day: number of days you want the image to stay OPTIONAL if 0, it will be available undefinitely
first-view: 1 OPTIONAL if not 0, the image will be deleted at first view

Jaex:
Scrxtchy did you found auto regex addition to URL field in latest version useful?

Jaex:
Added [lut.im](https://github.com/ShareX/ShareX/wiki/Custom-Uploader-examples#lutim) here.

PR #726, opened 2015-06-05
Add VideoBin support

corey-:
Add support for videobin.org as requested in #714

### DIFF
--- a/ShareX.UploadersLib/Enums.cs
+++ b/ShareX.UploadersLib/Enums.cs
@@ -124,7 +124,9 @@ namespace ShareX.UploadersLib
         Imgrush,
         SharedFolder, // Localized
         Email, // Localized
-        CustomFileUploader // Localized
+        CustomFileUploader, // Localized
+        [Description("VideoBin")]
+        VideoBin
     }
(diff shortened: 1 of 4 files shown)


Jaex:
Thanks.

Jaex:
I think your computer time is wrong, https://github.com/ShareX/ShareX/commits/master in here always showing your commit in top like this:

corey-:
Huh, sorry about that. Yes it appears my computer date got set a day ahead somehow.

k3d3:
@corey- this uploader doesn't work.

corey-:
@k3d3 I just tested it and it works fine. Videobin only accepts .ogv files (even though they say they'll convert other formats, haven't gotten them to work), so what you need to do is go Task Settings > Screen Recorder, then check the box that says "Run CLI afterwards:". Then click profiles, add a new profile. Make the "Path:" the path to ffmpeg.exe, "Arguments:" should be `-i %input %output`, and "Extension:" should be `ogv`.

k3d3:
ah. that should probably be specified somewhere.

PR #761, opened 2015-07-04
French Translation update

Shadorc:
Add missing translations, do some corrections.
I also added me as french translator, I hope you don't mind about it.


Jaex:
I don't mind. Thanks.

Issue #802, opened 2015-07-18
FTP file upload uses url-encode on file name instead of URL only


ElectronicWar:
When you upload a file with the custom FTP uploader that contains e.g. a Space or UTF-8 characters, ShareX will mangle the file to be url-encoded, uploads it and then generates a link with a lot of url encoded characters.

Example: `my file.txt` will be uploaded as `my%20file.txt` and the resulting link will be `http://myserver.tld/my%2520file.txt`.

It is especially frustating with UTF8 files, e.g. Japanese file names. It will make the filename completely unreadable.

maurits150:
This issue seems to have returned.

When I upload to SFTP using the following SubFolderPath: /...../data/img/%mon %y/, it actually creates a folder named July%202016 instead of using a normal space.

ElectronicWar:
This looks like it is "deeper" in the code, my fix was only for the final file path. If I find some time I will look at it, but this might take a few days.

PR #803, opened 2015-07-19
Do not url-encode file names for (S)FTP upload.

ElectronicWar:
Fix for Issue ShareX/ShareX#802
Do not url-encode file names for (S)FTP upload.

This is not necessary and a valid URL is created afterwards resulting in double-url-encoded URLs which is message and effectively destroys UTF8 based filenames e.g. Japanese.

### DIFF
--- a/ShareX.UploadersLib/FileUploaders/FTP.cs
+++ b/ShareX.UploadersLib/FileUploaders/FTP.cs
@@ -109,7 +109,6 @@ namespace ShareX.UploadersLib.FileUploaders
         {
             UploadResult result = new UploadResult();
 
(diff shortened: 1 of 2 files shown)


Jaex:
I hope you tested your change alot because everytime someone do ftp naming related commit somethings get fixed but other things get broken.

ElectronicWar:
I tested against a linux FTP and SFTP server (which are most common) with a pleora of strange filenames.
I do see no reason to url-encode filenames before upload other than catering to really old software. Which is probably still better to break than having destroyed file names for anything non-ASCII.

Issue #799, opened 2015-07-18
LnkU.co / 1EZ.link / dogr.net support


vavavr00m:
http://lnku.co/api/

Jaex:
@DanielMcAssey added lnku.co support and other domains pointing to that web site.

vavavr00m:
Thanks a bunch @DanielMcAssey !

PR #805, opened 2015-07-20
Adds LnkU.co URL Shortener

DanielMcAssey:
Issue #799 

Adds LnkU.co as a provider for URL shortening.

Tested and Works


Jaex:
Hi, thanks for it but I can't accept it when your entire changes using tab spaces while project using normal spaces when you check "Files changed" tab you can see none of the code indents matching.

https://github.com/ShareX/ShareX/wiki/Code-contribution

I wrote here tab spacing should use 4 spaces. Having half of codes using tab spacing others not is not acceptable.

DanielMcAssey:
Sorry about that, I completely missed the Contributing page on the wiki!

PR #822, opened 2015-07-24
Add extra flag to help with silent installations

PhilHannent:
When network administrators are deploying software en-mass they deploy using silent installations and the fact that ShareX opens a window when it has finished installing would be breaking that process.


Jaex:
How auto updater can open it then?

PR #843, opened 2015-08-01
Update French Translation

Shadorc:
Some corrections, add missing translations and remove those that no longer existed.

PS: We can translate "Disable hotkeys" but not "Enable hotkeys"


Jaex:
You updated it right before I was going to release it :D thanks.

Shadorc:
Yes, it's because I saw that it was no longer in beta that I finished the translation :p

Issue #863, opened 2015-08-04
Add Polr support


unixfox:
https://github.com/Cydrobolt/polr/wiki/Polr-Self-Hosted-API-Documentation

PR #865, opened 2015-08-04
Add Self-Hosted Polr support

DanielMcAssey:
Add Self-Hosted Polr support (https://project.polr.me/) as requested by #863 

Also support https://polr.me, however at the moment https://polr.me does not have a way to generate API keys, but it is planned, so I have added the possibility of using polr.me when user API keys are available, without further modification.


Jaex:
Thanks.

PR #907, opened 2015-08-14
VP9 on Screen recorder

Disinterpreter:
I make VP9(new WebM) on Screen recoder. 
Example: http://disinterpreter.me/img/lxjLpY6691439585806.webm

### DIFF
--- a/ShareX.ScreenCaptureLib/Enums.cs
+++ b/ShareX.ScreenCaptureLib/Enums.cs
@@ -78,6 +78,8 @@ namespace ShareX.ScreenCaptureLib
         libx264,
         [Description("VP8 (WebM)")]
         libvpx,
+        [Description("VP9 (WebM)")]
+        libvpxvp9,
         [Description("Xvid")]
         libxvid,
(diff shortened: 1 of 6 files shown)


Jaex:
I was tested it before and it was not working properly for real time encoding therefore I didn't added it. It is too early yet to add it.

Jaex:
I tried again now with latest ffmpeg and still frames dropping alot and when I stop, it stops 5 seconds later. So it is not suitable for real time encoding or require special commands for it.

Disinterpreter:
@Jaex i haven't droppings frames, but i have stopping on 5 second.

Jaex:
If you stopping on 5 seconds that means you dropping seconds of frames.

Disinterpreter:
@Jaex u can make VP9 on experemntal and write on combobox - "VP9 (experemental)"

Jaex:
When people see it have higher number majority of people would use it instead of VP8.

Disinterpreter:
@Jaex but "experemental" - unstable version

PR #927, opened 2015-08-18
Add 64-bit Architecture Support

### LINKED ISSUE #796: Issues with Registering DLL/OCX in 10.0.0 for Recorder Devices
I've ran into the same issue that was detailed here! https://github.com/ShareX/ShareX/issues/132
The issue that cause this was that the x86 version of the Visual C++ 2010 Redist libraries was downloaded onto my x64 system, and it tried to register the x64 DLL with the x86 version.

The "Recorder-devices-setup.exe" is failing to download the x64 version of Visual C++ 2010 Redist!

davidruhmann:
Fix #796

### DIFF
--- a/InnoSetup/Recorder devices setup.iss
+++ b/InnoSetup/Recorder devices setup.iss
@@ -5,6 +5,8 @@
 AppName={#AppName}
 AppVerName={#AppName} {#AppVersion}
 AppVersion={#AppVersion}
+ArchitecturesAllowed=x86 x64 ia64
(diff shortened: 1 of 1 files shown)


Jaex:
Thanks.
By the way I wonder why your commits adds new line to end of files always.

davidruhmann:
The newline is probably an artifact of using the online GitHub.com editor to make the change.  I did not add the newline myself.  Have to look into if that functionality can be disabled.

PR #929, opened 2015-08-19
Added keys for region selection #926

### LINKED ISSUE #926: Hotkey to simulate Mousedown on screen-capture screen.
Right now it requires that I Mousedown in order to initiate a screenshot, I'd like to be able to use a keyboard only option with capturing some smaller, pixel-sensitive elements on my screen without having to switch from the arrow keys, to the mouse for movement and then back to arrow keys. Not an awful experience on it's own, so this is merely a Quality-of-life type of enhancement.

In short, I'm targeting the [Hold Left Click] option and requesting that a keybind option (of some kind) be added if it hasn't already.

davidruhmann:
Created an implementation of this using the <kbd>Insert</kbd> and <kbd>End</kbd> keys.

__Quick mode__
* [Hold Left click] [Insert] Start region selection
* [Right click] [Insert] Cancel region selection
* [End] Stop region selection

__Multi region mode__
* [Hold Left click] [Insert] Start region selection
* [Right click] [Insert] Cancel region selection
* [End] Stop region selection
* [End] Stop editing the region

__Any Opinions?__
resolves #926


Jaex:
Good job :)

PR #930, opened 2015-08-19
multi direction arrow key movement

davidruhmann:
Allows two arrow keys to be pressed to allow diagonal movement during region selection.
__8 direction movement rather than just 4__


Jaex:
I don't think using System.Windows.Input and adding reference to PresentationCore and WindowsBase was necessary. You could add 4 different bool for 4 direction and keydown/up buttons will set true/false for these and have one function which will convert this bools to proper 8 direction movement.

davidruhmann:
That sounds like a good methodology for it as well. :+1:

Jaex:
Now it looks better, thanks.

PR #931, opened 2015-08-19
nameparser parameter support

### LINKED ISSUE #875: [request] Number of digits for incremental numbers (%i)
It would be a nice addition to be able to specify the number of digits desired for the incremental number tag in filenames.

Ex: setting it to 4 would generate: 0001, 0002, 0003, ...

davidruhmann:
resolve #875 

parameter = `{n}`
for entries i, ra, rn

`%i{n}` = 0 pad left by n
`%ra{n}` = repeat n times
`%rn{n}` = repeat n times


Jaex:
Another nice work :+1:

Jaex:
I noticed it not works when it is first characters:



But works fine when I add one character to beginning:

davidruhmann:
Ah, found the issue.  I was comparing the IndexOf result > 0 when it should be >= 0.

PR #936, opened 2015-08-20
change to insert and delete

davidruhmann:
related #926 

Changed from <kbd>Insert</kbd> and <kbd>End</kbd> to <kbd>Insert</kbd> and <kbd>Delete</kbd>

__Quick mode__
<kbd>Insert</kbd> - Start region selection
<kbd>Insert</kbd> - Stop region selection
<kbd>Delete</kbd> - Cancel region selection

__Multi region mode__
<kbd>Insert</kbd> - Start region selection
<kbd>Insert</kbd> - Stop region selection
<kbd>Insert</kbd> - Select/Deselect region
<kbd>Delete</kbd> - Cancel region selection

__Keyboard and Mouse switching scenarios__
<kbd>Left click</kbd> - Updated to work with <kbd>Insert</kbd> and selection starting and stopping

### DIFF
--- a/ShareX.ScreenCaptureLib/Forms/RectangleRegion.cs
+++ b/ShareX.ScreenCaptureLib/Forms/RectangleRegion.cs
@@ -348,7 +348,7 @@ namespace ShareX.ScreenCaptureLib
 
             if (AreaManager.IsCreating)
             {
-                sb.AppendLine(Resources.RectangleRegion_WriteTips__End__Stop_region_selection);
+                sb.AppendLine(Resources.RectangleRegion_WriteTips__Insert__Stop_region_selection);
                 sb.AppendLine(Resources.RectangleRegion_WriteTips__Right_click__Cancel_region_selection);
                 sb.AppendLine(Resources.RectangleRegion_WriteTips__Esc__Cancel_capture);
             }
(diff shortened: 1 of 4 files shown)


Jaex:
I found bug I think.

How to reproduce:

1. Hold right/left arrow.
2. Hold up/down arrow.
3. Release up/down arrow.
4. It not continue to move right/left and stopped.

davidruhmann:
From what I understand, this is due to how windows handles the key repeater messages.  When a new key (up or down) takes over the repeating, it does not go back to repeating the previous key (right or left) when the new key (up or down) is released.  I can look into working around this using a timer of sorts to process movement based on key state rather than key messages.

Jaex:
That will be ugly workaround.

davidruhmann:
@Jaex Here is an example emulation of the key repeater using timers: https://github.com/davidruhmann/ShareX/commit/9652457f74b52d80898f841419f2ffec30cde095
The performance was much better than I expected.  Seems to run at the same amount of CPU usage as the WM_KEYDOWN method from what I watched on the graph.  Let me know what you think.

Jaex:
@davidruhmann thanks for your efforts but codes already become very complicated and I don't think it is very big issue to make codes far more complicated to workaround it. I mean I didn't even understood what you did to solve it in your commit :( But I don't like timers running in background and always tried to avoid using timers in ShareX.

davidruhmann:
I understand not wanting to make it any more complicated.

Jaex:
Found slight issues now:

1. <kbd>Insert</kbd> to start.
2. <kbd>Right click</kbd> to cancel it.
3. <kbd>Left hold</kbd> not starts region selection.

davidruhmann:
__Ah, nice catch!__   The issue was that the IsCreating state needed to be reset for a new selection which is normally performed by the mouse up of the <kbd>left button</kbd>, but since the selection was started with <kbd>insert</kbd>, this was not happening.  Added the end selection routine to the <kbd>right button</kbd> mouse up after the cancel to simulate this action.  Had to wrap the end selection calls to prevent a region from being duplicated if the mouse was hovering over an existing region when cancelled.

PR #942, opened 2015-08-22
Correct a clear error

programmer-1:
$ symbol removed. it will break compile!


Jaex:
It is C# 6.0 feature you must be using old Visual Studio if it not compiles for you.

programmer-1:
yes You're right I use VS 2013, 
but in my opinion should be added to the description of the project.
thanks.

PR #950, opened 2015-08-27
resolve right click cancel to insert state

davidruhmann:
related #936 

The cancel operation does not reset the state back to be ready for a new region.  The end operation is needed to reset the state.

fixes issue:
<kbd>Insert</kbd> to start.
<kbd>Right click</kbd> to cancel it.
<kbd>Left hold</kbd> or <kbd>Insert</kbd> not starts region selection.
But If I cancel it with <kbd>Delete</kbd> then it works fine.


Jaex:
Thanks.

PR #952, opened 2015-08-27
disable minimize for color picker

davidruhmann:
The color picker if minimized can become stuck behind region annotate window.  I did not see a reason that minimizing is needed for this window.

When stuck you can escape by
* <kbd>Alt</kbd> + <kbd>Tab</kbd> to make color picker visible again
* <kbd>Esc</kbd> to close capture
* <kbd>Alt</kbd> + <kbd>F4</kbd> to close color picker
* <kbd>Win</kbd> then open back up the color picker from the task bar

Though, I would say that none of these are intuitive enough for an average user since there are no mouse only resolution.


Jaex:
Unfortunately I can't accept this because people can find minimize button useful for ```Tools -> Color picker``` window.

PR #955, opened 2015-08-28
add continuous integration support (appveyor and travis-ci)

davidruhmann:
Add support for the Continuous Integration services
* [__Travis-CI__](https://travis-ci.org) (Unix Mono xBuild)
* [__AppVeyor__](https://ci.appveyor.com) (Windows Visual Studio 2015 MSBuild)

both services free for open source projects

Reason for Changes
* Added `.travis.yml` and `appveyor.yml` configuration files
* xBuild with Mono does not support the (default) none property for DebugType
* Updated BuildEvents to work on Windows (cmd) and Unix (shell) systems

Examples
* https://ci.appveyor.com/project/davidruhmann/sharex
* https://travis-ci.org/davidruhmann/ShareX/builds

### DIFF
--- a/ShareX.UploadersLib/ShareX.UploadersLib.csproj
+++ b/ShareX.UploadersLib/ShareX.UploadersLib.csproj
@@ -57,7 +57,6 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
-    <DebugType>none</DebugType>
     <Optimize>true</Optimize>
     <OutputPath>bin\Release\</OutputPath>
     <DefineConstants>TRACE</DefineConstants>
@@ -795,11 +794,9 @@
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <Import Project="$(SolutionDir)\.nuget\NuGet.targets" Condition="Exists('$(SolutionDir)\.nuget\NuGet.targets')" />
   <PropertyGroup>
-    <PreBuildEvent>cd $(ProjectDir)APIKeys\
-
-if not exist APIKeysLocal.cs (
-    type nul &gt; APIKeysLocal.cs
-)</PreBuildEvent>
+    <PreBuildEvent>cd $(ProjectDir)APIKeys
+echo=&gt;&gt; APIKeysLocal.cs
+exit 0</PreBuildEvent>
   </PropertyGroup>
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
        Other similar extension points exist, see Microsoft.Common.targets.
--- a/ShareX.UploadersLib/ShareX.UploadersLib.csproj
+++ b/ShareX.UploadersLib/ShareX.UploadersLib.csproj
@@ -795,7 +795,7 @@
   <Import Project="$(SolutionDir)\.nuget\NuGet.targets" Condition="Exists('$(SolutionDir)\.nuget\NuGet.targets')" />
   <PropertyGroup>
     <PreBuildEvent>cd $(ProjectDir)APIKeys
-echo=&gt;&gt; APIKeysLocal.cs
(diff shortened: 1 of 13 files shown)


Jaex:
I was signed up to Travis-CI before but didn't want to add it because that compiled version won't be useable by users because it can't contain API keys. And when users can't even use it then what is point of using this services?

davidruhmann:
The main point of the service is to provide immediate build validation.  The compiled version from the ci builds do not have to be used.  That is my thoughts anyway.

Jaex:
Why would I commit something without build and test it? So when I commit something I know is it works or not works. Why would I open that page to check is it compiles when I can check it myself in Visual Studio. I don't get it. I mean compiling myself more quick way to test.

davidruhmann:
As long as you are the single gatekeeper that works, but whenever changes come from a pull request, this can provide validation without you having to first manually build it before committing.  Just trying to provide options for extra validation.

McoreD:
@davidruhmann thank you for this pull request; however, my immediate thoughts were the same as @Jaex because when we code in Visual Studio, we always build to see if it compiles properly. Even if we don't build, Visual Studio is intelligent enough to provide us with on-the-fly error messages if the solution is not compilable. 
Just saw your last message. Can you confirm that pull requests can be assessed to see if they compile before they are merged?

davidruhmann:
For example, that green check mark next to my second commit shows that it passed build on travis-ci.  It is the Build Pull Requests option that they provide.

ShareX.UploadersLib/ShareX.UploadersLib.csproj
```diff
@@ -795,11 +794,9 @@
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <Import Project="$(SolutionDir)\.nuget\NuGet.targets" Condition="Exists('$(SolutionDir)\.nuget\NuGet.targets')" />
   <PropertyGroup>
-    <PreBuildEvent>cd $(ProjectDir)APIKeys\
-
-if not exist APIKeysLocal.cs (
-    type nul &gt; APIKeysLocal.cs
-)</PreBuildEvent>
+    <PreBuildEvent>cd $(ProjectDir)APIKeys
+echo=&gt;&gt; APIKeysLocal.cs
```
McoreD: @davidruhmann is this checking if we already have a APIKeysLocal.cs as before? We don't want APIKeysLocal.cs overwritten. Thanks.

davidruhmann:
You would have to set it up for the actual ShareX/ShareX branch for it to check all pull requests.  That one just got checked since Travis-CI is running on davidruhmann/ShareX.

davidruhmann:
@McoreD  `echo=>> APIKeysLocal.cs` is only appending a new line to the file which will create the file if it does not exist or just add a new line if it does.  Not super elegant, but I can update it to have a file check if wanted.

ShareX.UploadersLib/ShareX.UploadersLib.csproj
davidruhmann: This is just trying to append (add to existing or create new if none exists) a new line to the APIKeysLocal.cs file.  I can add a file check if wanted.

ShareX/ShareX.csproj
```diff
@@ -1037,16 +1036,8 @@
...
-if $(ConfigurationName) == Release (
-    for %25%25x in (git.exe) do if not [%25%25~$PATH:x]==[] (
-        git rev-parse HEAD &gt; GitHash.txt
-    )
-)
-
-if not exist GitHash.txt (
-    type nul &gt; GitHash.txt
-)</PreBuildEvent>
+git &amp;&amp; git rev-parse HEAD &gt; GitHash.txt || echo= &gt;&gt; GitHash.txt
```
davidruhmann: This command is checking if git exists, if so, run the git command, else append a new line to the GitHash.txt file

Jaex:
Currently I'm not planning to use this because it don't have enough benefits to me. Sorry.

davidruhmann:
Understood, no need to be sorry. :smiley:
The PR will be here as a reference if you ever decide you want it.

PR #959, opened 2015-08-31
add base form with default icon

### LINKED ISSUE #958: ShareX icon missing from Alt+Tab menu
Many of the forms (monitor test, drag and drop, region...) in ShareX do not have an icon specified and thus do not display an icon in the <kbd>Alt</kbd> + <kbd>Tab</kbd> menu.

Side Node: What do you think about closing the surface selection forms when focus is lost?
* <kbd>Alt</kbd> + <kbd>Tab</kbd>
* <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>Escape</kbd>
* <kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>Delete</kbd>
* etc...

davidruhmann:
Added the default ShareX icon to a new BaseForm which other forms derive from to inherit the icon.  IMHO, the BaseForm helps with maintainability and provides a single location where form defaults can be specified without having to modify every form.

resolves #958

### DIFF
--- a/ShareX.HelpersLib/Automate/AutomateForm.cs
+++ b/ShareX.HelpersLib/Automate/AutomateForm.cs
@@ -34,7 +34,7 @@ using System.Windows.Forms;
 
 namespace ShareX.HelpersLib
 {
-    public partial class AutomateForm : Form
+    public partial class AutomateForm : BaseForm
     {
         private static AutomateForm instance;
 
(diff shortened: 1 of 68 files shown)


Jaex:
This is good idea. Also it could be used to add generic title like: "ShareX" will be default title and if you edit Title property then it will be appended to it like "ShareX - " + Title. That way when people translating they don't need to add "ShareX - " part theirself. But it is too late now because this can cause issues with previous translates. Also I'm not sure would this custom property show up in resx extension too.

Jaex:
I wonder why FTPClientForm.resx required change when other form don't require?

davidruhmann:
The others were not updated because I did not open the forms in the designer to cause the generated code to be updated. see #961

Jaex:
Unfortunately this change created far more work than what it solved to me:

https://dl.dropboxusercontent.com/u/14076298/ShareX/2015/12/tEtjqwFCmj.mp4

Every time I compile project while designer is open, it resets icon. So I must fix icon every single time :(

Issue #962, opened 2015-09-01
Send HTTP Headers in Custom Uploader


hpp3:
I want to create a custom uploader (using POST). I can edit the request body parameters, but I don't see any way to send an HTTP header. Specifically, I'm trying to implement a streamable.com uploader, which requires basic authentication to use their API. Basic authentication means I have to send HTTP headers in my request.

Jaex:
Main reason why I didn't add this because have no space in UI to put it.

hpp3:
I'm fine with it being a JSON only feature. The UI doesn't have to include a way to create headers. I would edit the JSON file to use this feature.

Jaex:
Implementing it would be like 2 minutes work but handling all that UI stuff hour of work. But I won't like making JSON only too. And I really can't find space in UI to put it.

hpp3:
Remove a few rows from the list of arguments. Then in that space you can add:
1) Edit JSON button. The button opens the JSON file and allows you to manually edit the JSON, to add the headers. Since it might be a rarely used feature, it might not be worthwhile to dedicate UI room for headers specifically. Making the feature JSON only and then putting a button to edit JSON in the UI is a middle ground.
or
2) A single header input field (label "HTTP Header" + 2 text boxes, for key and value).
or 
3) A field to perform authentication. A label "Authentication" + 2 text boxes for username and password. If there's no more room, then it will do basic authentication. If there's more room, then a dropdown selection of authentication methods.

atillabyte:
This should work in the new pull @hpp3 - I didn't take your UI into consideration so I tried to sort it out myself. I took a screenshot of it [here] (https://url.atil.la/lerga).

Jaex:
I found better way to fit headers to UI:

![](https://dl.dropboxusercontent.com/u/14076298/ShareX/2015/10/6jDJsVMj0F.png)

Jaex:
There is one problem now. I can't test this headers feature because I don't have any custom uploader which works with headers.

PR #1021, opened 2015-09-30
CustomUploader Headers & CustomUploader UI

atillabyte:
Related to Issue #962


Jaex:
Sorry I don't liked your solution.

PR #1053, opened 2015-10-08
Fixed dropfile deletion url

dominik-th:


Jaex:
I was reported this to them:


But they still didn't fix it.

Your fix uses api to request deletion not opens deletion page.

nshtg:
@Jaex 
Since they still did not fix it, could you consider the original PR? It's frustrating to open and change the link every time.

The behaviour would be consistent with the one you see with Up1 etc and would match what I expect when it says "Open deletion URL".

PR #1061, opened 2015-10-13
Added Seafile Support

zikeji:
I tested as much as I could find, for further testing others are open to downloading the experimental setup with Seafile support.

__Download:__ https://seacloud.cc/f/71f961cd4c/


Jaex:
Destination settings window not shows ShareX icon for some reason in your commit. When I return to previous commit it shows properly. But can't find what is wrong.

zikeji:
Odd. I am done for the night I will look into it tomorrow (I am in UTC-5,
2AM for me right now).  Roll it back for now,  sorry I didn't notice.

Jaex:
I just reverted your resx changes and indeed resx was causing it. This gonna be very complicated to fix too because can't even see full comparison. I wonder now is there anything else broken in resx :(

zikeji:
All I did was add the Seafile icon to the resx. Could Git have corrupted
the resx? Maybe restore old and add Seafile image again.

Jaex:
I found way to see all differences just add ".patch" to end of commit link.
And from there I searched "Icon" and indeed you changed form icon there.

zikeji:
I didn't change form icon, at least not intentionally. Sorry i caused you
trouble.

Jaex:
I meant your commit changed.

Jaex:
This is example ss from Seafile app I think:



More obvious there what URL to use but when I used that URL showed there validate button gave error. So still dunno what URL to use.

zikeji:
Use auto complete URL.  Should he /api2 on the end of the URL.  I didn't
want to auto add it in case the private server version people have hosted
uses a different end URL. Should be https://seacloud.cc/api2

ShareX.UploadersLib/Forms/UploadersConfigForm.Designer.cs
```diff
@@ -2527,6 +2574,267 @@ private void InitializeComponent()
...
+            this.txtSeafileUploadLocationRefresh.Click += new System.EventHandler(this.txtSeafileUploadLocationRefresh_Click);
+            // 
+            // lblSeafileSelectLibrary
+            // 
+            resources.ApplyResources(this.lblSeafileSelectLibrary, "lblSeafileSelectLibrary");
+            this.lblSeafileSelectLibrary.Name = "lblSeafileSelectLibrary";
+            // 
+            // grpSeafileAccInfo
+            // 
+            this.grpSeafileAccInfo.Controls.Add(this.myListView1);
```
Jaex: I can't find this in UI.

zikeji: I was gonna say I didn't know what those were.  I should have looked harder
for remnants like that.

Jaex:
Because you reading and replying my messages from email I bet you don't see my countless edits I keep doing after send messages.

Gonna use dropdown control for API URL now.

ShareX.UploadersLib/Forms/UploadersConfigForm.cs
```diff
@@ -2044,6 +2060,252 @@ private void txtPomfResultURL_TextChanged(object sender, EventArgs e)
...
+            Seafile sf = new Seafile(txtSeafileAPIURL.Text, txtSeafileAuthToken.Text, null);
+            Seafile.SeafileCheckAccInfoResponse SeafileCheckAccInfoResponse = sf.GetAccountInfo();
+
+            if (SeafileCheckAccInfoResponse == null)
+            {
+                MessageBox.Show(Resources.UploadersConfigForm_Login_failed, "ShareX", MessageBoxButtons.OK, MessageBoxIcon.Error);
+                return;
+            }
+            txtSeafileAccInfoEmail.Text = SeafileCheckAccInfoResponse.email;
+            txtSeafileAccInfoUsage.Text = HelpersLib.NumberExtensions.ToSizeString(SeafileCheckAccInfoResponse.usage).ToString() + " / " + HelpersLib.NumberExtensions.ToSizeString(SeafileCheckAccInfoResponse.total).ToString();
```
Jaex: I don't understand why you used extensions like that in here. Not to mention you converted string to string there.
Extensions used like this: ```SeafileCheckAccInfoResponse.usage.ToSizeString()```

zikeji:
I saw your past edit but not the ones before (just read them).  Sorry.  And
I used auto complete feature on the one control, guess it wasn't intuitive
enough.

ShareX.UploadersLib/Forms/UploadersConfigForm.cs
```diff
@@ -2044,6 +2060,252 @@ private void txtPomfResultURL_TextChanged(object sender, EventArgs e)
...
+            {
+                if (SeafileLibrary.permission == "rw")
+                {
+                    ListViewItem libraryItem = lvSeafileLibraries.Items.Add(SeafileLibrary.name);
+                    libraryItem.Name = SeafileLibrary.id;
+                    libraryItem.Tag = SeafileLibrary;
+                    libraryItem.SubItems.Add(HelpersLib.NumberExtensions.ToSizeString(SeafileLibrary.size));
+                    if (SeafileLibrary.encrypted)
+                    {
+                        ListViewItem.ListViewSubItem isEncrypt = libraryItem.SubItems.Add("\u221A");
```
Jaex: This sub item not used at all.

ShareX.UploadersLib/Forms/UploadersConfigForm.cs
```diff
@@ -558,6 +559,21 @@ public void LoadSettings()
...
             txtPomfResultURL.Text = Config.PomfUploader.ResultURL;
 
+            // Seafile
+
+            txtSeafileAPIURL.Text = Config.SeafileAPIURL;
+            txtSeafileAuthToken.Text = Config.SeafileAuthToken;
+            txtSeafileDirectoryPath.Text = Config.SeafilePath;
+            txtSeafileLibraryPassword.Text = Config.SeafileEncryptedLibraryPassword;
+            txtSeafileLibraryPassword.ReadOnly = (Config.SeafileIsLibraryEncrypted ? true : false);
+            btnSeafileLibraryPasswordValidate.Enabled = (Config.SeafileIsLibraryEncrypted ? false : true);
```
Jaex: ReadOnly one not make sense here and for validate one you could have just used !Config.SeafileIsLibraryEncrypted

zikeji: That sub item is for encrypted folders.  Try adding an encrypted folder to
your account and you will see what I'm talking about.

Jaex:
This looks better:



By the way it would be far better if we used messenger instead of spamming here :P

ShareX.UploadersLib/Forms/UploadersConfigForm.cs
Jaex: isEncrypt don't used anywhere.

zikeji:
It does look better like that!  Sorry I haven't used  C# in a couple months
so some of the shortcuts are rusty. Been into Golang working on a
lightweight URL shortener.

And yeah messenger could work but I'm in bed replying to these emails from
my phone lol. 2:40 am

Jaex:
Why days to expiration is 7 by default and I can't make it never expire?

Jaex:
Upload is successful: https://seacloud.cc/f/b7ed3e3d63/
Good job!

Now I must try to find way to use current Uploader methods for ShareFile method because it is currently very ugly.

Jaex:
I wonder is it possible to get this direct link https://seacloud.cc/f/b7ed3e3d63/?raw=1 redirected from it.

Jaex:
I fixed accept header issue it was not working properly when set in header. But still always getting:

The remote server returned an error: (415) UNSUPPORTED MEDIA TYPE.

Response:
{"detail": "Unsupported media type '' in request."}

zikeji:
>Edit: Finally made it work. This API is ridiculous.

That's exactly what I was thinking. Why use PUT for a request that could have been handled easily with GET, and why the heck return variables in the header and not in JSON?!

Jaex:
Did you missed my question?

zikeji:
> Did you missed my question?

Could you repeat the question?

Jaex:
Jaex commented 3 days ago
   Why days to expiration is 7 by default and I can't make it never expire?

zikeji:
>Why days to expiration is 7 by default and I can't make it never expire?

Setting it to 0 should set it to never expire. The API didn't have enough info on the matter though, the 7 I just left as an arbitrary number.

Jaex:
You can't set it to 0. Minimum is 1 you did. Also by default why anyone would want it to expire.

Jaex:
By the way are you using Steam? For to be able to easily communicate.

zikeji:
> You can't set it to 0. Minimum is 1 you did. Also by default why anyone would want it to expire.
And if you make it possible to set 0 then you must still mention that 0 disables it. Not everyone can guess it.

Should probably be an option to disable it then.

BTW, expiration only applies to the shared link. If you upload a file to share, you generally don't want the link accessible indefinitely. The expiration doesn't effect the file at all, just the public share link.

PR #1064, opened 2015-10-14
Update Lambda uploader

mstojcevich:
Updated to use the new upload API, updated the favicon, fixed broken links, and changed text in the config page to be accurate


Jaex:
Few people was requesting option to use lambda.sx url instead of λ.pw url.

mstojcevich:
I was thinking that too, I'll probably add that soon

Issue #1008, opened 2015-09-23
[Suggestion] ReOpen Post capture tasks dialog after failed upload/save


Slind14:
Hey,

I have come across this problem multiple times now and in some cases it isn't possible to get the image back. Would it be possible to reopen the after capture dialog in case the upload/save of a picture failed?

Jaex:
How save fails?

Slind14:
NAS in restart for example.

PR #1094, opened 2015-10-31
Reopen save as dialog on errors after capture

mpistrich:
Reopens the save dialog after the exception window has been confirmed. This fix should prevent losing pictures on errors (see issue #1008).

An easy way to provoke an error on saving:

1. Create drive mapping to a folder: ``subst S: C:\Test``.
2. Activate post capture task ``Save image to file as...``.
3. Create breakpoint before the image is saved.
4. Capture a image and select the drive ``S:`` as location.
5. While the program is stopped, unmap the drive via ``subst S: /D``.
6. Remove breakpoint, continue execution. An exception is shown.
7. After the exception the file dialog is shown again.

### DIFF
--- a/ShareX/ImageData.cs
+++ b/ShareX/ImageData.cs
@@ -24,8 +24,10 @@
 #endregion License Information (GPL v3)
 
 using ShareX.HelpersLib;
+using ShareX.Properties;
 using System;
 using System.IO;
+using System.Windows.Forms;
 
 namespace ShareX
 {
@@ -46,6 +48,9 @@ namespace ShareX
             catch (Exception e)
             {
                 DebugHelper.WriteException(e);
+                MessageBox.Show(new Form() { WindowState = FormWindowState.Maximized, TopMost = true },
+                    string.Format(Resources.ImageData_Write_Error + "\r\n\r\n" + e, filePath), "ShareX",
+                    MessageBoxButtons.OK, MessageBoxIcon.Error);
             }
 
             return string.Empty;
--- a/ShareX/Properties/Resources.Designer.cs
+++ b/ShareX/Properties/Resources.Designer.cs
@@ -1036,6 +1036,15 @@ namespace ShareX.Properties {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to Could not write image to path {0}..
+        /// </summary>
+        public static string ImageData_Write_Error {
+            get {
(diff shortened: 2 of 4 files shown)


Jaex:
After image save failed if you re open dialog people won't know why it keep re opening. User will keep trying same path and think there is error in ShareX. You need to show proper error message box to let users know about it.

mpistrich:
Currently it displays a message box which is created by ``Helpers.CreateDirectoryIfNotExist`` before the save to file dialog is reopened. The message box looks like this: ErrorMessageBox

Do you think this is sufficient? If not the Helpers method must be adapted to now always show a message box and a more detailed and user friendly error message must be shown.

Jaex:
I thought your errors was happening when trying to create file, not directory.

mpistrich:
You are right, no message box was shown when problems concerning the file occurred. I've added a message box in the ``ImageData.Write`` method, but I am not that happy with this change. I think message boxes should be shown from some helper class (e.g. ``DebugHelper``) but that's a topic for another change.

ShareX/ImageData.cs
```diff
@@ -46,6 +48,9 @@ public string Write(string filePath)
             catch (Exception e)
             {
                 DebugHelper.WriteException(e);
+                MessageBox.Show(new Form() { WindowState = FormWindowState.Maximized, TopMost = true },
```
Jaex: I'm not sure why you did new Form() here.

Shadorc:
Hi ! 

I don't know if it's a desired behavior (but i think not) but in post-capture tasks, I checked :  "Save as ..." and "Upload" and before the update, I could cancel the save and the upload was still done but, now, cancel the save also cancels the upload.

I think this pull-request is the cause because it was the only change related to "Save as..."

abhiagile:
Thanks Jaex.

Exact same issue which Shadorc has reported.. Any fix?

Jaex:
It is fixed now.

abhiagile:
For me it is still the same, do we have to update anything or add?

abhiagile:
I removed sharex and downloaded again from your website but 10.4 still gives the same error..

Do I have to do anything else?

Jaex:
I meant fixed in codes.

abhiagile:
I see so you mean i need to find a file and replace the code?

Can u give me the steps to do it and also the code.. it would be great help!

Jaex:
If you don't know how to compile yourself then you need to wait next version.

Issue #846, opened 2015-08-02
Using "Disable/Enable hotkey" to disable hotkeys disables some keys.


vavavr00m:
I set CTRL + ALT + H as my "disable/enable hotkeys" hotkey and SHIFT + F and SHIFT + D for other tasks. After disabling ShareX hotkeys with CTRL + ALT + H, I cannot use SHIFT + F and SHIFT + D to quickly capitalize the two letters. Just tried to set CTRL + A to do a certain task and after disabling ShareX with the Disable/Enable hotkey, I cannot use CTRL + A (highlight all) anymore.

Jaex:
Where you can't use ctrl + a to highlight all?

vavavr00m:
Chrome (while typing here) & Notepad so far.

Jaex:
Reason is actually I don't unregister hotkey but make them not trigger their tasks. Main reason for this is if i disable whole hotkey system then "disable hotkey" hotkey won't work too. So yes I did workaround  like this. And I'm lazy to make proper unregistering system :P

vavavr00m:
I understand :-D Please consider this issue a request ("Enhancement") then ^_^

PR #1095, opened 2015-10-31
Issue #846: Unregister hotkeys when disabling them.

mpistrich:
Hotkeys are unregistered temporarily until the user enables them again. The hotkey form now also reacts to enabling and disabling hotkeys. When hotkeys are disabled, a new hotkey is saved but not registered. Resetting hotkeys to default also enables them again.


Jaex:
Thanks.

PR #1110, opened 2015-11-12
pomf.io changes

neko:
made changes to pomf.io related uploader


Jaex:
Your changes will break it.
Adding this not valid: "https://u.pomf.io" because response already includes full URL.

PR #1179, opened 2015-12-07
Updated to .NET framework 4.5

andreassjoberg:
Updated to .NET framework 4.5


Jaex:
.NET framework 4.5 not supports Windows XP.

andreassjoberg:
Ok. There's some incompatibility with Newtonsoft.Json. I get errors when compiling. When using an older version (6.0.8), the application doesn't manage to save my settings. Some exception is thrown in SaveAsync when debugging.

Kesmy:
I'm also not able to build as of the addition of the MegaApiClient package, for this reason.

The primary reference "MegaApiClient, Version=1.1.3.98, Culture=neutral, PublicKeyToken=0480d311efbeb4e2, processorArchitecture=MSIL" could not be resolved because it has an indirect dependency on the assembly "Newtonsoft.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed" which was built against the ".NETFramework,Version=v4.5" framework. This is a higher version than the currently targeted framework ".NETFramework,Version=v4.0"

Jaex:
@gpailler are you have any idea why this issue could be happening? Few people also contacted me, they couldn't compile ShareX because of this problem and I can't figure out why:



Because this problem happens to very few people.

johnduhart:
@Jaex I ran into that issue as well, what's happening is the compiler is picking up on a version of Newtonsoft.Json that's installed into the GAC or in the reference path. For me, it was a copy that was part of the Azure SDK.

gpailler:
Thanks @johnduhart for the explanation. I will try to reproduce the build error and will try to fix it with a BindingRedirect in app.config

gpailler:
So it's a known issue with AzureSDK 2.3. There is lot on complaints on Internet about that and I cannot fix it on MegaApiClient side. But I plan to switch NewtonSoft.Json version for MegaApiClient to an higher version and it will probably fix this error.

In the meantime, users can follow these links to solve the issue:
http://blogs.visoftinc.com/2014/08/10/dll-hell-strikes-again/
http://stackoverflow.com/questions/24922637/visual-studio-is-throwing-a-wrong-compile-time-exception

PR #1190, opened 2015-12-12
Adds s-ul support 2

corin12355:
Non 6 month old version.
Adds support for https://s-ul.eu


ShareX.UploadersLib/FileUploaders/Sul.cs
```diff
@@ -0,0 +1,99 @@
...
+            APIKey = apiKey;
+        }
+
+        public override UploadResult Upload(Stream stream, string fileName)
+        {
+
+            Dictionary<string, string> args = new Dictionary<string, string>();
+            args.Add("wizard", "true");
+            args.Add("key", APIKey);
+            args.Add("client", "sharex-native");
```
Jaex: What is this for? [User-agent](https://en.wikipedia.org/wiki/User_agent) header already shows that uploader is using ShareX.

Jaex:
Thanks.

ShareX.UploadersLib/FileUploaders/Sul.cs
corin12355: It was supposed to be renamed to "client-native" as it enables errors to be sent as json rather than just showing the error straight up. If another client wanted to take advantage of it they could do. I'll leave it as it stands for now though and will update it later on.

PR #1203, opened 2015-12-22
Fixed transparency for clipboard upload

oxysoft:
`Clipboard.GetImage()` discards the alpha channel and turns all transparent pixels to black. I committed some extra code that will handle 24bpp and keep alpha transparency when uploading.

note: the preview window still shows black and should implement these new changes as well to display transparent pixels, perhaps moving this code into an utility function and use that rather than Clipboard.GetImage to avoid code duplication.

### DIFF
--- a/ShareX/UploadManager.cs
+++ b/ShareX/UploadManager.cs
@@ -29,8 +29,10 @@ using ShareX.Properties;
 using ShareX.UploadersLib;
 using System;
 using System.Drawing;
+using System.Drawing.Imaging;
 using System.IO;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Web;
 using System.Windows.Forms;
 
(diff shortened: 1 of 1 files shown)


Jaex:
Very high percentage of software only copy 24 bit bitmap with no transparency info. How you found app which can copy with transparency? Even many apps copy png image just to have transparency support.

oxysoft:
PyxelEdit (pixel art oriented software) is the software I used to test this commit and works well with it

Jaex:
Your code assumes clipboard always have dib but what if it only have https://msdn.microsoft.com/en-us/library/system.windows.forms.dataformats.bitmap(v=vs.110).aspx then it will fail.

Did you tried copying transparent image from browsers or other image editors? I wonder is any of them support 32 bit copying.

Jaex:
Also you can copy your method to here:
https://github.com/ShareX/ShareX/blob/master/ShareX.HelpersLib/Helpers/ClipboardHelpers.cs

Jaex:
I can't try PyxelEdit because it is not free app.

oxysoft:
I just tried this with Paint.net and the transparent pixels are indeed turned to white which is weird because if I then paste it into paint.net, the transparency is obviously kept. This makes me think that Paint.net must be adding some additional information and whatnot to the clipboard data.

And if that wasn't enough, transparent pixels are turned to black with Gimp.

Jaex:
Okey I will merge and do that changes myself.

Jaex:
I did few changes to DIB codes and also added PNG support that way there is more chance that images will keep transparency. And moved them to ClipboardHelpers class.

Jaex:
I was right about that DIB codes not looks stable. It shifts right side of image to left like mentioned in here: #1261

PR #1211, opened 2015-12-28
changed pomf.io to pomf.is

neko:
I have moved pomf.io to pomf.is.


neko:
i'm not sure whether this pull request will affect any current pomf.io users, as i've set it up so that people uploading to pomf.io can still upload their stuff.

if you believe it'll interfere with current users, reject the pull request.

it's also worth noting that pomf.pl has merged into pomf.io, but removing it would likely affect current users.

Issue #1265, opened 2016-01-18
Please add an After Capture Task: Show in Explorer


moniuch:
Rather than digging up the current screenshot folder with right click, I'd rather have the saved screenshot file revealed automatically. Thanks.

junderhill:
This would definitly be a worth while enhancement. I may try and implement this if I get chance.

moniuch:
Hi,

Thanks for recognizing the demand.

It would also be helpful to have a global system key combo for opening the
current screenshot folder. It is difficult to do so from within an external
shortcut utility because the folder name changes each month.

Regards and thanks for your work,
Artur

Jaex:
There is already hotkey for it:

![](https://dl.dropboxusercontent.com/u/14076298/ShareX/2016/01/ShareX_zSplkHurNT.png)

moniuch:
Good call on the screenshot folder task. Haven't realized one could add
that.
I expected though that after defining such a task it would appear as an
option in the list of after capture tasks. This is however not the case. So
how can I now use the task as After Capture part?

PR #1270, opened 2016-01-19
Added post capture task to show in explorer #1265

junderhill:
I've implemented the functionality discussed in #1265. This has been added as an After Capture task that can be enabled. If enabled it shows windows explorer after capture with the new screenshot highlighted. See screenshots below.

explorer_2016-01-19_23-10-10

I've added the localized translation for English for 'Show file in explorer'. I wasn't sure how you do the translations for the other resource files?

Please let me know if there are any issues.
Thanks

### DIFF
--- a/ShareX.HelpersLib/Properties/Resources.Designer.cs
+++ b/ShareX.HelpersLib/Properties/Resources.Designer.cs
@@ -197,6 +197,15 @@ namespace ShareX.HelpersLib.Properties {
         }
         
         /// <summary>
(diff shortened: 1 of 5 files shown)


Jaex:
After capture menu there follows order of execution and you added it last when it was executing before uploading. Also there is already method in Helpers class for open file in explorer.

junderhill:
@Jaex I've updated my branch based on your comments above. Thanks

Issue #1273, opened 2016-01-20
Return HTTPS link for Imgur upload


karmiphuc:
Imgur does support HTTPS (https://imgur.com, for example https://i.imgur.com/JmUGY7D.png), so it's safer to return HTTPS link for Imgur upload.

Jaex:
#716

PR #1274, opened 2016-01-20
Using security protocol for return url

daofresh:


daofresh:
solve #1273

Jaex:
#989 #689

PR #1297, opened 2016-02-02
Allow for silent installation

strawgate:
Installer currently launches sharex after installation regardless of /silent or /verysilent command line arguments


Jaex:
#822

PR #1303, opened 2016-02-04
#1298 fix

### LINKED ISSUE #1298: someimage API does not give direct url link
this can be hacked together, but their api explicitly avoids giving a direct link

image id is consistent from their 'direct' link to the actual image file

a `http://someimage.com/asdf` direct link leads to a direct image of `https://i1.someimage.com/asdf.filetype` in every file ive uploaded. but it's kind of unsafe to assume they'll always use `i1` as subdomain

I like this host because I havent hit a filesize cap :p

jevanpipitone:
#1298 fix


Jaex:
Having empty api key by default can cause issues to current users.

jevanpipitone:
Done what you specified.
Direct URL option defaults to true presently for someimage, or should it default to false?

ShareX.UploadersLib/Forms/UploadersConfigForm.Designer.cs
```diff
@@ -4040,165 +4228,17 @@ private void InitializeComponent()
...
             // actRapidShareAccountType
             // 
             resources.ApplyResources(this.actRapidShareAccountType, "actRapidShareAccountType");
             this.actRapidShareAccountType.Name = "actRapidShareAccountType";
             this.actRapidShareAccountType.SelectedAccountType = ShareX.UploadersLib.AccountType.Anonymous;
             // 
+            // label1
+            // 
+            resources.ApplyResources(this.label1, "label1");
+            this.label1.Name = "label1";
```
Jaex: I don't like having unnamed controls in forms.

Jaex:
By the way if that labels tells you, that you can get API key from somewhere. Instead you could use link label control which would open API page when clicked on it.

Jaex:
You could write something for me to know you updated it (that way i would receive email notification).

Jaex:
I noticed now your resx changes broken majority of control sizes :(
Are you using higher dpi Windows?
This is how it is now: 
This is how it was before:

Jaex:
Because in 20000+ lines of changes I can't figure out whats wrong or not it is not possible for me to fix it manually therefore I'm forced to revert this commits unfortunately.

PR #1308, opened 2016-02-04
Marked X and Y label as @Invariant

dannoe:


Jaex:
Most of the things not marked as invariant because everytime I edit something in designer those fields getting removed so it is useless to re add them because it will be removed again. I only re add them before release.

PR #1323, opened 2016-02-14
#1298

jevanpipitone:
I think I've fixed the problem with the resx now.


Jaex:
In your direct link you adding .jpg extension when it can be .png like this:
https://i1.someimage.com/OJdYywZ.png
why?

PR #1375, opened 2016-03-01
IRC Client not in the ShareX anymore.

LoliHook:
Removed it from Readme.


Jaex:
Most of things outdated in README. Changing one line not makes difference. I need to update all of changes.

PR #1381, opened 2016-03-04
Handle multiple context menu selections - Fixes #770

### LINKED ISSUE #770: Context menu uploading doesn't add the files in order and duplicates files in list
Number of files to be uploaded: 24
Image

Queued files:
Image in queue

Mega error:
Mega

campbeb:
The problem is that multiple context menu selections spawn multiple .exe
(invoked). Each invoked .exe sets the `InstanceProxy` variables in the
running .exe and tells it to start a thread to process. But, there is no
waiting, so the `InstanceProxy` variables get clobbered before the spawned
thread can process them.
This fix uses a semaphore to control the invoked .exe. There is only one
entry allowed forcing invoked .exe to wait until any other invoked .exe
is finished running and releases the semaphore.

Fixes #770

### DIFF
--- a/ShareX.HelpersLib/SingleInstanceApplication/ApplicationInstanceManager.cs
+++ b/ShareX.HelpersLib/SingleInstanceApplication/ApplicationInstanceManager.cs
@@ -34,10 +34,13 @@ namespace SingleInstanceApplication
 {
     public static class ApplicationInstanceManager
     {
+        private static Semaphore semaphore;
+
         [DebuggerStepThrough]
         public static bool CreateSingleInstance(string name, EventHandler<InstanceCallbackEventArgs> callback, string[] args)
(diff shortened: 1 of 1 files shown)


Jaex:
Thanks.

Jaex:
I did few testings using these files:



Most of the times few files was duplicate in ShareX main window list:



So your solution is not reliable too unfortunately.

campbeb:
Can you verify that your context menu was invoking your test build and not some other copy of the exe on your computer?

Jaex:
Yes it was my test build. Without test build majority of them becomes same file.

campbeb:
I apologize for the incomplete solution. It worked on my computer.
I've made a change and will open a new PR.

PR #1382, opened 2016-03-04
Handle multiple context menu selections - Move semaphore release

campbeb:
Current process is as follows:
1. Invoked .exe gets the semaphore
2. Invoked .exe sets `InstanceProxy` values in running exe
3. Invoked .exe causes thread to spawn in running exe
4. New thread in running .exe reads values from `InstanceProxy`
5. Invoked .exe releases the semaphore

The problem is that there is no guarantee on how fast new threads in
running .exe will spawn and read values from `InstanceProxy`. This PR
addresses that by moving the semaphore release from the invoked .exe to
the thread spawned in running .exe.

This means that an invoked .exe can't change `InstanceProxy` until 
running .exe has processed the previous call.

### DIFF
--- a/ShareX.HelpersLib/SingleInstanceApplication/ApplicationInstanceManager.cs
+++ b/ShareX.HelpersLib/SingleInstanceApplication/ApplicationInstanceManager.cs
@@ -54,8 +54,6 @@ namespace SingleInstanceApplication
                     UpdateRemoteObject(name);
 
                     if (eventWaitHandle != null) eventWaitHandle.Set();
-
-                    semaphore.Release();
                 }
 
                 Environment.Exit(0);
@@ -125,6 +123,8 @@ namespace SingleInstanceApplication
             if (callback == null) return;
 
             callback(state, new InstanceCallbackEventArgs(InstanceProxy.IsFirstInstance, InstanceProxy.CommandLineArgs));
+
+            semaphore.Release();
         }
     }
 }


Jaex:
Tested many times now and it worked perfectly :)

campbeb:
Great! I'd hate to mess it up twice.

Jaex:
Writing this to change log so complicated last 10 minutes trying figure out how to write it :D 

* Now only one CLI command will be transfered to running ShareX process in same time to handle multiple file uploading from Windows Explorer, so when Windows Explorer try to send multiple files to ShareX process, those will be reached to ShareX one by one in sync (by [@campbeb](https://github.com/campbeb))

campbeb:
That sounds really complicated.

Maybe

- Windows Explorer context menu "Upload with ShareX" now correctly handles multiple files

That makes it understandable without having to get all technical with users.

Jaex:
Not just "Upload with ShareX" but also it applies to "Send to" menu too.

campbeb:
- Windows Explorer context menus "Upload with ShareX" and "Send To ShareX" now correctly handle multiple files

Also, in case you hadn't seen it, if ShareX isn't running, bad things will still happen (multiple copies will launch). But, this is in the SingleInstance code, not the command line handling. I think the SingleInstance stuff doesn't happen soon enough, so multiple copies get launched. You might get some new bug reports if people try testing the bug fix.

Jaex:
Then it is better to not write it to change log.

campbeb:
Probably a good idea. 

Looking at the single instance code is on my list of things to look at some time.

PR #1394, opened 2016-03-09
Add Japanese translate.

netman1221:


Jaex:
It would be better if you can pull request all translation files together instead of one by one.

PR #1396, opened 2016-03-10
Proper Single Instance Implementation

campbeb:
Reworked single instance code to better handle being launched with
context menu.

Previously:
- On launch, process would try to open an `EventWaitHandle`. If it
didn't exist, then process assumed it was first and created it
- Multiple processes launching at same time would all see the
`EventWaitHandle` did not exist before any process could create it
- This led to multiple instances of program running (as discussed at https://github.com/ShareX/ShareX/pull/1382#issuecomment-194561241)

Now a mutex is used for proper synchronization:
- Processes will all try to open the mutex. Only first one will succeed
- Process that opens the mutex will setup the `EventWaitHandle`
- Other processes won't get the mutex and will be directed to use the
`EventWaitHandle` to pass their command line args to the running
instance

Notes:
 - Currently, there is a 100ms wait for the mutex before it times out.
This works fine for my testing, but might need to be higher if there
are problems when invoking from context menu and ShareX isn't 
already running
 - Everything works fine in my testing, but this will need to be more 
widely tested. Any problems should only be seen when using context
menu to Upload or Send To or using `-multi` command line parameter

### DIFF
--- a/ShareX.HelpersLib/SingleInstanceApplication/ApplicationInstanceManager.cs
+++ b/ShareX.HelpersLib/SingleInstanceApplication/ApplicationInstanceManager.cs
@@ -35,56 +35,45 @@ namespace SingleInstanceApplication
     public static class ApplicationInstanceManager
     {
         private static Semaphore semaphore;
+        private static string appName = "ShareX";
+        private static string eventName = string.Format("{0}-{1}", Environment.MachineName, appName);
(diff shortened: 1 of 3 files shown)


Jaex:
Codes for single instance become so complicated with so many sync mechanism I can't even understand how it works anymore. So I can't be sure about is something can go wrong.

campbeb:
Yeah, I was afraid it might look bad. Hopefully the PR comment at least gives you an idea of what is going on.

PR #1424, opened 2016-03-20
[WIP] Refactoring uploaders using Inversion of Control

johnduhart:
This is an early, work in progress, refactor of something that's always bugged me about ShareX's codebase. Right now to add an additional image uploader, it requires edits to for different locations within the codebase. This also results in tight coupling, and makes automated testing difficult. I'm submitting this PR to gather feedback on the design and overall concept (the code is sloppy and will be fixed in due time).

I've refactored a few uploaders in a way that allows for inversion of control, and introduced a service locator in the `WorkerTask` for retreiving these new implementations, rather than the existing switch statement. I've also added Simple Injector as a dependency injection framework.

### `IUploadService`
Represents an upload service such as pastebin or Dropbox. The interface provides metadata about the service, and can creates `IUploader`s and `IUploadServiceConfig`. Upload-type descendents exist such as `IURLShortenerService` and `ITextUploadService`

### `IUploader`
Mentioned above, the `IUploader` implementation actually handles the task of uploading things to the service.

### `IUploadServiceFactory`
Code that needs to locate an instance of an `IUploadService` will make use of the factory to find an `IUploadService` by ID or enum value

### `IUploadServiceConfig`
**Naming sucks, I need to change it**

Handles UI concerns of displaying the upload target in the `UploadersConfigForm`. Rather than having all of the code for each uploader UI defined in the config form, the `IUploadServiceConfig` would provide an icon and UserComponent for that uploader.

## Wrap-up
It's 3AM here, so I'll answer any questions tomorrow morning, and continue to update this PR.

### DIFF
--- a/ShareX.HelpersLib/ServiceLocator.cs
+++ b/ShareX.HelpersLib/ServiceLocator.cs
@@ -1,0 +1,27 @@
+﻿using System;
+using SimpleInjector;
+
+namespace ShareX.HelpersLib
+{
+    public static class ServiceLocator
+    {
+        private static Container Container;
+
+        public static void SetContainer(Container container)
+        {
+            Container = container;
+        }
+
+        public static TService GetInstance<TService>() where TService : class
+        {
+            EnsureContainerIsSet();
+            return Container.GetInstance<TService>();
+        }
+
+        private static void EnsureContainerIsSet()
+        {
+            if (Container == null)
+                throw new InvalidOperationException(@"ServiceLocator used before a container was set");
+        }
+    }
+}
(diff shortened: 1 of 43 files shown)


McoreD:
Thank you for this pull request. A similar implementation is now live in master without the need for any 3rd party libraries . We welcome you to check out this latest code to seek your feedback.

johnduhart:
I provided feedback to @Jaex directly, but one of the things that is still missing from the implementation is Dependency Injection. It does make some good strides in this regard but it can still be improved on. I wrote a summary of what IOC is in general here: https://gist.github.com/johnduhart/4e135b22c179a95fe9a4

If the issue is an additional library, [TinyIoC](https://github.com/grumpydev/TinyIoC) can solve this. It included using a single .cs file in your solution.

johnduhart:
I'm closing this PR for the time being.

PR #1429, opened 2016-03-21
Ukrainian language file

6c6c6:
Ukrainian translation


Jaex:
It would be better if you can pull request all translation files together instead of one by one.

6c6c6:
Ok

PR #1435, opened 2016-03-23
Refactored sharing services to use new pattern

johnduhart:
I've updated the sharing services to use the new uploader patterns. I've also moved the validation helpers outside of the `UploadersConfig`.


Jaex:
Thanks.

PR #1438, opened 2016-03-23
Add verified result URLs to Pomf uploaders

ErikHumphrey:


Jaex:
Really? I made them automatic so even when result urls change it will work. And you adding manual ones.

ErikHumphrey:
Ah, the manual Result URL box was disabled (greyed out) when I tried, so I had to change the source code. Nonetheless I still got a working fix for one of them referenced in #1439

Jaex:
Result URL box never becomes disabled what you mean?

ErikHumphrey:
never becomes enabled I guess? I wasn't able to input my own result URL, but relaunching sharex fixed the problem so I guess it's a non-issue

PR #1439, opened 2016-03-23
Add cocaine.ninja result URL

ErikHumphrey:
cocaine.ninja does not appear to have an automatic result URL and has an 100% fail rate (ShareX doesn't have toast notifications anymore, so I don't really know what the error was, I just heard the little blip) despite the site being fine; I compiled this and now it has an 100% success rate


Jaex:
I can't upload to that website with your changes too, it keep giving this error: `The underlying connection was closed: An unexpected error occurred on a send.` So there is problem in their end.

PR #1453, opened 2016-03-26
ServerCertificateValidationCallback is a property, not an event.

johnduhart:
Store the previous value and then restore it.


Jaex:
In your code if two uploader tried to upload in same time then previousValue can be value of ServerCertificateValidationCallback and unable to reset to actual previousValue.

PR #1491, opened 2016-04-08
Created ImageLoader for loading preview images in a new thread

TheBoJaN:


Jaex:
Your code not supports loading url, also i suspect it supports cancellation too, also don't liked your code sorry.

TheBoJaN:
You're right, I just assumed ImageHelpers.LoadImage handled URLs as well, that's an easy fix. What do you mean by "i suspect it supports cancellation too"? Give me some feedback and I can improve on it. 

I'm not sure what I can do about you not liking how my code works however... It simply moves image loading out of the main thread and closes the loading thread if it receives no activity within 10 seconds to avoid having an idle thread sitting around all day.

PR #1502, opened 2016-04-17
P/Invoke fixes

dbremner:
This patch does four things:
1. Add CharSet.Unicode to DllImports that lack it - CharSet.Ansi is the [default](https://blogs.msdn.microsoft.com/oldnewthing/20140812-00/?p=263).
2. Fixes the [SendMessage](http://www.pinvoke.net/default.aspx/user32.sendmessage) and [SendMessageTimeout](http://www.pinvoke.net/default.aspx/user32.SendMessageTimeout) P/Invokes for 64bit and adds the needed casts to callers.
3. Fixes the return type of [GetWindowLong](http://www.pinvoke.net/default.aspx/user32/GetWindowLong.html). This function returns a uint but was declared as returning a ulong.
4. Add CharSet.Unicode to a struct.

### DIFF
--- a/ShareX.HelpersLib/Extensions/Extensions.cs
+++ b/ShareX.HelpersLib/Extensions/Extensions.cs
@@ -189,12 +189,12 @@ namespace ShareX.HelpersLib
 
         public static void BeginUpdate(this RichTextBox rtb)
         {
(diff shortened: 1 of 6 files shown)


Jaex:
Was ShareX giving any error or incorrect behavior without your changes?

Also why you added CharSet.Unicode to few things which don't even use string anywhere.

PR #1503, opened 2016-04-17
Add pomf.lesderid.net to Pomf.cs

lesderid:


Jaex:
Why you added `https://pomf.lesderid.net/f`? Your response already contains full url. It is only required for people who using old version of pomf which not returns full url.

lesderid:
Whoops, I didn't notice the difference. I couldn't actually test it since I'm not on Windows. I fixed it and added my clone to the readme.

Jaex:
Don't add anything to read me. Otherwise people will see it and wonder why it is not exist in current release. Read me supposed to update after release not before.

lesderid:
Makes sense. Removed it from the readme.

PR #1517, opened 2016-04-24
Update cuntflaps.me

alucard0134:
uses ssl by default.


Jaex:
Look like that website now returns full url in response, can you remove `"https://a.cuntflaps.me"` parameter.

PR #1534, opened 2016-05-03
Added an option to get direct links from Google Drive uploads.

### LINKED ISSUE #1522: Google Drive Direct Link
It is possible to generate direct URL from response.
https://developers.google.com/drive/v2/reference/files#resource-representations
`downloadUrl`
`embedLink`

And also, as a last resort: http://www.labnol.org/internet/direct-links-for-google-drive/28356/
I'm pretty sure the response does include the file ID .

dannoe:
fixes #1522


Jaex:
`webContentLink` is not direct link otherwise I would have added it.

dannoe:
webContentLink
string
A link for downloading the content of the file in a browser using cookie based authentication. In cases where the content is shared publicly, the content can be downloaded without any credentials.

Source: https://developers.google.com/drive/v2/reference/files

PR #1539, opened 2016-05-06
French Translation updated

Spidersouris:


Jaex:
I don't think InnoSetup uses other than english in ShareX setup file.

PR #1558, opened 2016-05-16
Updated URL's and URL handling.

lithium720:
Added a domain, and the 'i' sub domain configurations.
Changed to building the URL locally.


ShareX.UploadersLib/FileUploaders/Lithiio.cs
```diff
@@ -77,7 +76,7 @@ public override UploadResult Upload(Stream stream, string fileName)
 
             if (result.IsSuccess)
             {
-                result.URL = result.Response;
+                result.URL = Config.UploadURL + result.Response;
```
Jaex: Tip: You can use `URLHelpers.CombineURL(Config.UploadURL, result.Response);` to combine urls more safely.

PR #1564, opened 2016-05-18
Add support for context menu keyboard key to uploads listview

stdray:
subj


Jaex:
What is "uploads listview" means.

stdray:
Control lvUploads on MainForm.
sharex vshost_2016-05-18_19-04-12

PR #1618, opened 2016-06-11
Save last drawing mode in Greenshot

### LINKED ISSUE #1609: Option to save the last used tool in image editor
Hi! I'm a new fan of ShareX. The way i use it mostly is this: region capture -> image editor -> highlight something in the image (via 'Draw line' or 'Draw arrow' tool mostly) -> image upload to imgur. 

Everything works cool and fine, except the default tool is always 'Draw rectangle' and there is _no way_ to change that or arrange tools on the panel (i checked ShareX settings, image editor settings, even the .ini file with a lot of settings for the image editor in the ShareX directory). Image editor saves other user stuff like line colors, line thickness, font settings etc., but not the tool itself sadly. 

The last 11.0.1 update mentions "SaveSettingsAfterTaskCompleted" setting, which "saves all settings after each task completed", but the tool selection in the image editor is still not saved, /sadpanda. 

Please consider adding the option (or saving the last used tool by default even?)

campbeb:
Saves the last drawing mode used in Greenshot when editor is closed so
that it will be selected next time image editor is opened.
Closes #1609

### DIFF
--- a/Greenshot.ImageEditor/Forms/ImageEditorForm.cs
+++ b/Greenshot.ImageEditor/Forms/ImageEditorForm.cs
@@ -192,7 +192,7 @@ namespace Greenshot
                 SurfaceSizeChanged(Surface, null);
 
                 BindFieldControls();
-                _surface.DrawingMode = DrawingModes.Rect;
+                _surface.DrawingMode = EditorConfiguration.RememberLastDrawingMode ? EditorConfiguration.LastDrawingMode : EditorConfiguration.DefaultDrawingMode;
                 RefreshEditorControls();
                 // Fix title
                 if (_surface != null && _surface.CaptureDetails != null && _surface.CaptureDetails.Title != null)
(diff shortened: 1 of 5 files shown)


Greenshot.ImageEditor/Forms/ImageEditorForm.cs
```diff
@@ -1011,7 +1013,8 @@ private void RefreshEditorControls()
             FieldAggregator props = _surface.FieldAggregator;
             // if a confirmable element is selected, we must disable most of the controls
             // since we demand confirmation or cancel for confirmable element
-            if (props.HasFieldValue(FieldType.FLAGS) && ((FieldType.Flag)props.GetFieldValue(FieldType.FLAGS) & FieldType.Flag.CONFIRMABLE) == FieldType.Flag.CONFIRMABLE)
+            if (props.HasFieldValue(FieldType.FLAGS) && ((FieldType.Flag)props.GetFieldValue(FieldType.FLAGS) & FieldType.Flag.CONFIRMABLE) == FieldType.Flag.CONFIRMABLE
+                && _surface.HasSelectedElements) // if nothing is selected, there is nothing to cancel, so don't disable controls
```
campbeb: This is necessary to overcome bug if Crop is the last drawing mode. When editor is opened back up, all tools on left are disabled, because it thinks we are in the middle of cropping.

PR #1623, opened 2016-06-11
Add "Use direct link" setting to Google Drive uploader

mihe:
I was trying to find some way to get around the fact that Google Drive encodes videos after uploading them from ShareX, which can take quite a while. All I want is a direct link to the uploaded file.

One way to create a direct link is to just append the ID to a URL, as such:

`$"https://drive.google.com/uc?id={upload.id}"`.

... but I figured it would be better to leverage the response somehow, so instead I tried using the `webContentLink` entry from `result.Response`, which contains the above mentioned link.

But by default `webContentLink` is bundled with the query `?export=download`, which seems to add `Content-Disposition: attachment` to the HTTP headers, forcing the browser to download the file instead of just viewing it. Stripping away said query parameter seems to fix that, and the resulting link can be viewed in the browser just fine.

Let me know what your thoughts are.

Also, thanks for an awesome application.

### DIFF
--- a/ShareX.UploadersLib/Forms/UploadersConfigForm.Designer.cs
+++ b/ShareX.UploadersLib/Forms/UploadersConfigForm.Designer.cs
@@ -4986,6 +4986,6 @@
         private System.Windows.Forms.TextBox txtEmailAutomaticSendTo;
         private System.Windows.Forms.CheckBox cbEmailAutomaticSend;
         private System.Windows.Forms.Button btnLithiioGetAPIKey;
-		private System.Windows.Forms.CheckBox cbGoogleDriveDirectLink;
-	}
+        private System.Windows.Forms.CheckBox cbGoogleDriveDirectLink;
+    }
 }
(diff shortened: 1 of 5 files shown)


ShareX.UploadersLib/Forms/UploadersConfigForm.Designer.cs
```diff
@@ -4977,5 +4986,6 @@ private void InitializeComponent()
         private System.Windows.Forms.TextBox txtEmailAutomaticSendTo;
         private System.Windows.Forms.CheckBox cbEmailAutomaticSend;
         private System.Windows.Forms.Button btnLithiioGetAPIKey;
-    }
+		private System.Windows.Forms.CheckBox cbGoogleDriveDirectLink;
```
Jaex: I wonder why tab characters used here.

Jaex:
It is indeed gives direct link, nice find.

mihe:
There, replaced the tabs with spaces, my bad.

PR #1641, opened 2016-06-16
Changed Imgur generated links to SSL

SpyTec:
Seeing as Imgur does support SSL links, I strongly prefer the links to automatically be https links rather than http.

Small change, just a quick change in how image links are given to user


Jaex:
#689

PR #1650, opened 2016-06-20
Added Azure Storage as File Upload Provider

0xdeafcafe:
Added a new file provider - Microsoft Azure.


Jaex:
Unfortunately I don't like using external libraries. As you can see there is countless uploaders in ShareX and I could easily use each of these services official .NET libraries (Google Drive, OneDrive, Dropbox etc.) but still decided to write everything myself. I mean if I used external libraries in each uploader then ShareX would have like 40x uploader library. Your Azure implementation requires 6 different libraries, so it is not acceptable for me. Sorry.

0xdeafcafe:
I assumed this would be acceptable due to the fact that the Amazon S3 uses external libraries, although that does only use 2 compared to 6. I'll see about writing a custom wrapper for Azure Storage and update this pull request.

Jaex:
Amazon S3 was implemented without external library here: #83
But later to support new regions of Amazon he instead decided to use external library here: #636
and even I didn't like using external library had no choice other than accepting it to support new regions :(

So if first implementation in #83 used external library I wouldn't have accepted it too.

dannoe:
@Jaex  What is the exact reason for this? 
Not using existing libraries is an anti-pattern of software development:
see [Not-invented-here Syndrom (wikipedia)](https://en.wikipedia.org/wiki/Not_invented_here#In_computing) or [Anti-Pattern (wikipedia)](https://en.wikipedia.org/wiki/Anti-pattern#Methodological)

0xdeafcafe:
Ah, that makes sense then. Like I said, I'll look into writing a small wrapper of the Http Api that Microsoft exposes and update this pull request accordingly.

@dannoe that is true, but for applications like this, the file size of it due to having so many dependencies would outweighs the pros.

Ideally, it would be good to actually take all the uploaders out of ShareX and have a sort of modular marketplace where you could just download the uploaders you want and have them hook in. That way you could keep using third party libraries and keep file sizes as low as possible.

dannoe:
@0xdeafcafe You're right. But if I am right, he's against 3rd-party libs in general.

McoreD:
Modular idea is good. We did discuss this possibility; however, it is still in the drawing board because, in a hypothetical scenario where I have downloaded all the uploaders modules (say *.Uploader.dll files), I would have over 50 dlls loading in ShareX runtime which could slow down startup time significantly.

dannoe:
You don't have to load them all at startup. You could lazy load them at use.

Jaex:
@dannoe:

If I didn't have good reasons I wouldn't be against using external libraries for uploaders.

* Each uploader can require many libraries which all of them using their own uploading classes etc. which makes using same uploading settings difficult like how ShareX sets user agent, sets proxy, reports progress with exact same info, error reporting method and all other web request settings used. Not in all libraries you have full control on it.

* Even single uploader library can be more than 1mb and having 50x uploader it can easily exceed 50mb when ShareX smaller than 5mb.

* Most important reason is startup time will slow alot. No you can't lazy load them. ShareX will have reference to setting files exist in most of libraries to be able to load settings. Not to mention other things can be required other than settings while loading too. And because these uploader libraries mainly huge it will be far more slower to load them in memory because msil doing compiling in background.

So both performance, size and quality of software can decrease greatly. I could be always lazy and use built in libraries too but instead I worked hard and implemented all uploaders myself. So I expect same thing from pull requests too.

dannoe:
I agree with your arguments.
The lazy loading thing was for the modular idea from McoreD not for referencing external libs.

PR #1656, opened 2016-06-22
Added pomf.gocataclysm.com

Ninn0x4F:


Jaex:
Comma is in wrong place, previous line should have it not last line.

PR #1677, opened 2016-06-28
Files and folders are now opened in the default filemanager.

### LINKED ISSUE #1673: [Suggestion] "Show file in explorer" opens folder despite the fact it's already opened
Hello. I'm using last version of ShareX (11.0.1) on Windows 8.
When we are using "Show file in explorer" as "After capture task" the folder with captures is opened every time i made a screeshot, even if it is already opened.

When I'm making 3 screenshots — 3 folder windows are opened. Etc.

Suggestion: check whether a folder with captures already opened and don't open it again.

dannoe:
Also fixes #1673.

### DIFF
--- a/ShareX.HelpersLib/Helpers/Helpers.cs
+++ b/ShareX.HelpersLib/Helpers/Helpers.cs
@@ -417,7 +417,7 @@ namespace ShareX.HelpersLib
             {
                 try
                 {
-                    Process.Start("explorer.exe", folderPath);
+                    Process.Start(folderPath);
                     return true;
                 }
                 catch (Exception e)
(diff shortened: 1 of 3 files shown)


Jaex:
There is problem in this approach: `Process.Start(folderPath);`

When folder path is this: `D:\Dev\GitHub\ShareX\ShareX\ShareX\bin\Debug\ShareX`

It actually running: `D:\Dev\GitHub\ShareX\ShareX\ShareX\bin\Debug\ShareX.exe`

Which causing this issue: #1950

Jaex:
As solution I'm planning to do this:

`folderPath = folderPath.TrimEnd('\\') + '\\';`

Which adds \ to end of path if path not contains it already. That way it will be explicit that it is folder path.

dannoe:
I would prefer this:

```
if (!folderPath.EndsWith(@"\"))
    folderPath = folderPath + @"\";
```

So you don't have to manipulate the string if it's already correct.

PR #1678, opened 2016-06-28
Added confirmation dialogs to two "Reset" buttons.

dannoe:
Are the names for the localization variables okay?


Jaex:
Need to add "..." to these buttons because there is dialog now.

Issue #1720, opened 2016-07-18
Change default Pomf clone to a decent clone.


neko:
Pomf.cat's uptime the last few weeks has been really unreliable; I'd recommend using one of the more reliable clones from [this spreadsheet](https://docs.google.com/spreadsheets/d/1kh1TZdtyX7UlRd55OBxf7DB-JGj2rsfWckI0FPQRYhE/edit#gid=0) as the default.

I'd probably suggest mixtape.moe myself.

Compare pomf clone uptimes [here](http://pomfstatus.sapph.io/).

banksymate:
Having the other owner change the NS back to CloudFlare since I'm out of town, our cache server is going down way too often and we've been aware and have been trying to address it but it hasn't been working out so I'm moving back to CloudFlare.

PR #1723, opened 2016-07-18
Change Pomf Default to more reliable host

Drybones5:
To fix issue #1720 (https://github.com/ShareX/ShareX/issues/1720).

Pomf.cat's uptime according to http://pomfstatus.sapph.io/ is abysmal and shouldn't be enabled as the default Pomf uploader host destination.

### DIFF
--- a/ShareX.UploadersLib/UploadersConfig.cs
+++ b/ShareX.UploadersLib/UploadersConfig.cs
@@ -264,7 +264,7 @@ namespace ShareX.UploadersLib
 
         // Pomf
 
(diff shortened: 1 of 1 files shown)


banksymate:
Sorry, I've been out of town and ever since I moved from CloudFlare (they've been causing me issues), the cache server has been going down constantly and I've been nagging the host about it. Haven't been able to do much since I'm gone but I'll try to move back to CloudFlare for now, it'll cause me a lot better uptime than I've been getting.

PR #1747, opened 2016-07-26
README.md parity with latest Pomf domain changes

ErikHumphrey:
Chevereto is also out-of-date but I don't use chevereto (yet) so I can't be bothered to put the new additions on that list


Jaex:
README should be updated fully not partially. I mean it should be updated version by version bases to be easily manageable. So current README is in specific version and I can check next versions in changelog to fill it.

PR #1779, opened 2016-08-12
Update Pomf.cs

nerd:


Jaex:
Your Pomf already returns full url, so no need to write `https://dev.null.vg`.

PR #1826, opened 2016-08-25
Sort destinations by name

Efreak:
I'm not entirely sure if this will work or not--I haven't actually tried compiling (because I don't want to download the nuget dependencies), but I'm hoping that this will sort the destinations as listed in the context menu (and in the main program as well).

I was recently told that puush was added, and couldn't find it because I was looking under pomf, which I use. It seems that maybe some of the services are currently sorted by popularity (twitter/imgur/pastebin at the tops of their lists), but there's simply too many services for that now, especially for the file uploader.

It would be even nicer to have an option to hide each service from the list, or to automatically hide unconfigured services that don't allow anonymous login.

### DIFF
--- a/ShareX.UploadersLib/Enums.cs
+++ b/ShareX.UploadersLib/Enums.cs
@@ -30,30 +30,30 @@ namespace ShareX.UploadersLib
     [Description("Image uploaders"), DefaultValue(Imgur)]
     public enum ImageDestination
     {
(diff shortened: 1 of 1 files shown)


Jaex:
I don't prefer destinations to be sorted by name.

KamranMackey:
@Jaex Maybe add an option to sort destinations by name then?

PR #1841, opened 2016-08-26
Fixes #1838

### LINKED ISSUE #1838: Video Thumbnailer
6b2a10d4-8d4d-481e-bb14-f8750feb446f

dannoe:


Jaex:
How exactly this can fix his problem? I'm not sure what was his problem too.

dannoe:
You are setting the `DefaultOutputDirectory` on the `ToolsSettings `but inject the `ToolsSettingsReference` into the `VideoThumbnailerForm`.

Jaex:
I already saw it now. Still not explains why it was null.

dannoe:
Because the `DefaultOutputDirectory` is set on the wrong object instance?

Jaex:
It is clone of reference object so it must be identical. But I'm too busy currently can't check codes.

dannoe:
No, there are two different instances:
devenv_2016-08-27_00-08-03

`taskSettings.TaskSettingsReference` holds a reference to the `Program.Default` instance.
but `taskSettings` itself, does get a copy of Program.DefaultTaskSettings

Jaex:
Now I checked codes and saw what is wrong, sorry.

By the way now appveyor every commit will compile setup automatically that way who have their issues fixed can try it instantly.

PR #1863, opened 2016-09-01
Accessibility: Keyboard Navigation

Paul-C-F:
Dialogs containing OK, Cancel and Close buttons can be dismissed by
hitting Enter or Escape on the keyboard, depending on the buttons
available.
Also added a Close button to the ChromeForm window.

### DIFF
--- a/ShareX.HelpersLib/Colors/ColorPickerForm.Designer.cs
+++ b/ShareX.HelpersLib/Colors/ColorPickerForm.Designer.cs
@@ -419,6 +419,7 @@
             // 
             // ColorPickerForm
             // 
+            this.AcceptButton = this.btnOK;
             this.AutoScaleMode = System.Windows.Forms.AutoScaleMode.None;
             this.BackColor = System.Drawing.SystemColors.Window;
             this.CancelButton = this.btnCancel;
--- a/ShareX.HelpersLib/Colors/GradientPickerForm.Designer.cs
+++ b/ShareX.HelpersLib/Colors/GradientPickerForm.Designer.cs
@@ -112,6 +112,7 @@
             // 
             // btnCancel
             // 
+            this.btnCancel.DialogResult = System.Windows.Forms.DialogResult.Cancel;
             resources.ApplyResources(this.btnCancel, "btnCancel");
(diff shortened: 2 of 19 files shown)


Jaex:
In color picker you need to press enter to confirm number so color can update, same for gradient form too. Probably there is many other such misuses too. Pressing enter in random control not always means closing form.

Paul-C-F:
@Jaex I didn't modify the ScreenColorPicker but the ColorPickerForm which has an OK button.

Are you willing to review it before just closing?

I will go over the screens one more time to make sure that the Accept behavior is OK on each one of them, the Escape behavior however is MUCH needed everywhere there's a cancel button.

Jaex:
Are you willing to review it before just closing?

What is that even means? I pulled your pull request tested it and then closed it.

Paul-C-F:
@Jaex Did you really test it cause there is NO accept button on the screen color picker form.

Just forget it, you have attitude and I don't want to deal with it anymore, enjoy your project.

Jaex:
I'm talking about color picker form not screen color picker form.

Who have attitude here implying that I'm telling lie to you...

Jaex:
Color changes when you press <kbd>Enter</kbd> in numeric up down control:
https://dl.dropboxusercontent.com/u/14076298/ShareX/2016/09/psTelywue1.mp4

Paul-C-F:
@Jaex You were right about that, I never tested the spinner fields.

But here's one trick you didn't know, the TextChanged event is not on the property inspector but can be assigned via code, I tested and it works perfectly, way better than pressing Enter.

Now the question is, even if that was not possible, would you discard all other changes only because of the color picker form?

`public ColorPickerForm(Color currentColor)
        {
            InitializeComponent();
            Icon = ShareXResources.Icon;

            SetCurrentColor(currentColor, false);
            nudRed.TextChanged += new EventHandler(colorNum_TextChanged);
            nudGreen.TextChanged += new EventHandler(colorNum_TextChanged);
            nudBlue.TextChanged += new EventHandler(colorNum_TextChanged);
        }

        private void colorNum_TextChanged(object sender, EventArgs e)
        {
            RGB_ValueChanged(sender, e);
        }`

PR #1899, opened 2016-09-11
Closes #1466 - Uplea Integration

### LINKED ISSUE #1466: Support for Uplea.
Can you add support for Uplea ? http://uplea.com/api

osfancy:
The Uplea API requires the file payload to be submitted as part of the POST parameter collection. To leverage existing code in the Uploader had to make a few modifications there. Biggest change is a marker in the arguments dictionary value if the value should be converted from a Base64 encoded string to a byte array. If the marker isn't present then the Uploader should continue behave as it did before. Let me know if I may have missed any code which might have precluded this change to the Uploader.


Jaex:
http://uplea.com/api I don't see here anything about mention of base 64. Also look like you submitted file two times. One as base 64 and also as bytes, if I understood correctly. So it feels like something wrong.

Jaex:
In UploadData method you using "file" as file form name but it should be "files[]".

osfancy:
@Jaex I think I follow - let me take another look at it, perhaps that was the issue I was running in to earlier.

Jaex:
When they meant post parameters it was just means post request. So just using usual UploadData method will be sufficient without require changes in Uploader.cs. Because UploadData method already using post parameter for file uploading.

osfancy:
@Jaex Thanks for the feedback, let me undo those changes and get it working without a change to Uploader.cs before I submit another PR.

osfancy:
@Jaex Yup I had misunderstood how that argument was being used :) I've reverted the change to Uploader and confirmed it's working correctly using "files[]".

Jaex:
This is concerning too in resx file:

`14,306 additions, 8,339 deletions not shown because the diff is too large. Please use a local Git client to view these changes.`

Also I saw you using `DataContractJsonSerializer` but ShareX have reference to Json .NET library.

osfancy:
I can update it to use Json .NET and can take a look at the resx changes - should have only added an icon.

Jaex:
You must not store UploadersConfig inside uploader. UpleaUploaderService.CreateUploader is responsible for filling uploader setting properties using UploadersConfig.

Jaex:
`upleaUserInformation.EmailAddress = (string)upleaGetUserInformationResponseObj.SelectToken("result.mail");`

Json .NET can automatically convert json to object but you doing them manually one by one for each property. Check this example: `JsonConvert.DeserializeObject<DropboxAccount>(response);`

osfancy:
@Jaex Thanks for the direction. I've replaced the use of DataContractJsonSerializer and in its place used JsonConvert. I've also updated the constructor to only pull out the API key when the config object is passed to it. Taking a look now at why the resx file is showing so many differences.

Jaex:
`[JsonProperty(PropertyName = "password")]`

You don't need to use JsonProperty, Json.NET is smart enough to ignore case.

Jaex:
`public Uplea(UploadersConfig config)`

Don't give uploaders config here. Make API key public property and set it inside `GenericUploader CreateUploader(...)` method.

Jaex:
`upleaGetUserInformationResponse.Status == bool.TrueString.ToLower()`

Just declare Status as bool not string.

Jaex:
+                upleaUserInformation.EmailAddress = upleaGetUserInformationResponse.Result.EmailAddress;
 +                upleaUserInformation.IsPremiumMember = upleaGetUserInformationResponse.Result.IsPremiumMember;
 +                upleaUserInformation.InstantDownloadEnabled = upleaGetUserInformationResponse.Result.InstantDownloadEnabled;

Why you declared this class two times when you can return Result directly.

osfancy:
@Jaex Agreed on bool and the duplicate class. With the duplicate class I was trying to maintain separation of public versus private objects but do see how it is unnecessary and isn't adding value. I've removed it. I've updated CreateUploader to set the public API key property based on the config value - I had misunderstood your suggestion previously. As far as setting the property name on the JSON attributes in the Response objects it was really just to be explicit and consistent as I needed them on the Request objects. This is because the Uplea API balks at "Username" and "Password" (which is what serialization without those attributes would produce. If we submit "Username" and "Password" the API response indicates that it doesn't see that a username/password was supplied. Their system is apparently case sensitive.

osfancy:
@Jaex I've committed an update to UploadersConfigForm.resx to resolve the wild number of differences previously presenting in the diff for that file. Turns out that this was due to VS adding in duplicates to the file and shuffling everything around. I had to de-dupe the new resx, sort it (and the baseline), then pick out the differences and append them to the original (unsorted) baseline version.

ShareX.UploadersLib/Forms/UploadersConfigForm.Designer.cs
```diff
@@ -412,6 +381,20 @@ private void InitializeComponent()
...
+            this.gbUpleaLoginCredentials = new System.Windows.Forms.GroupBox();
+            this.btnUpleaLogin = new System.Windows.Forms.Button();
+            this.lblUpleaApiKey = new System.Windows.Forms.Label();
+            this.txtUpleaApiKey = new System.Windows.Forms.TextBox();
+            this.lblUpleaPassword = new System.Windows.Forms.Label();
+            this.lblUpleaUsername = new System.Windows.Forms.Label();
+            this.txtUpleaPassword = new System.Windows.Forms.TextBox();
+            this.txtUpleaUsername = new System.Windows.Forms.TextBox();
+            this.gbUpleaUserInformation = new System.Windows.Forms.GroupBox();
+            this.label4 = new System.Windows.Forms.Label();
```
Jaex: All controls should have name.

ShareX.UploadersLib/Forms/UploadersConfigForm.resx
```diff
@@ -3762,119 +3762,119 @@ store.book[0].title</value>
   </data>
   <data name="pbDropboxLogo.Image" type="System.Drawing.Bitmap, System.Drawing" mimetype="application/x-microsoft.net.object.bytearray.base64">
     <value>
-        iVBORw0KGgoAAAANSUhEUgAAAOcAAAA8CAYAAACdDa3MAAAABGdBTUEAALGPC/xhBQAAACBjSFJNAAB6
```
Jaex: Why there is change here?

ShareX.UploadersLib/FileUploaders/Uplea.cs
```diff
@@ -0,0 +1,231 @@
...
+            }
+
+            return string.Empty;
+        }
+
+        private string GetUpleaResponse<T>(T upleaRequest) where T : IUpleaRequest
+        {
+            Dictionary<string, string> parameters = new Dictionary<string, string>();
+            parameters.Add("json", JsonConvert.SerializeObject(upleaRequest));
+            return SendRequestURLEncoded(upleaBaseUrl + upleaRequest.RequestType, parameters);
```
Jaex: Did you tried SendRequestJSON method?

ShareX.UploadersLib/FileUploaders/Uplea.cs
```diff
@@ -0,0 +1,231 @@
...
+        }
+
+        public override UploadResult Upload(Stream stream, string fileName)
+        {
+            var upleaBestNode = GetBestNode();
+
+            Dictionary<string, string> args = new Dictionary<string, string>();
+            args.Add("api_key", UpleaApiKey);
+            args.Add("token", upleaBestNode.Token);
+            args.Add("file_id[]", string.Format("{0}", Guid.NewGuid()));
```
Jaex: Why string format here?

ShareX.UploadersLib/FileUploaders/Uplea.cs
```diff
@@ -0,0 +1,231 @@
...
+
+            UploadResult result = UploadData(stream, string.Format("http://{0}/", upleaBestNode.Name), fileName, "files[]", args, contentType: "multipart/form-data");
```
Jaex: It is already "multipart/form-data"

ShareX.UploadersLib/FileUploaders/Uplea.cs
```diff
@@ -0,0 +1,231 @@
...
+            var uploadResult = JsonConvert.DeserializeObject<UpleaGetUpleaUploadResponse>(result.Response);
+
+            result.IsURLExpected = true;
```
Jaex: This is true by default.

ShareX.UploadersLib/Forms/UploadersConfigForm.resx
osfancy: The diff is due to formatting, the content should be the same.

ShareX.UploadersLib/FileUploaders/Uplea.cs
osfancy: Previously was including more than just the Guid. Also, don't have to call ToString on NewGuid when using string.Format. I've removed string.Format here now.

ShareX.UploadersLib/FileUploaders/Uplea.cs
osfancy: Removing redundant argument.

ShareX.UploadersLib/FileUploaders/Uplea.cs
osfancy: Removing set on IsURLExpected.

ShareX.UploadersLib/Forms/UploadersConfigForm.Designer.cs
osfancy: Will add missing names.

ShareX.UploadersLib/FileUploaders/Uplea.cs
osfancy: Yes, unfortunately SendRequestJSON didn't work for this. This API is very picky. I couldn't get it to respond successfully when using application/json as the Content-Type. GetUpleaResponse and consequently SendRequestURLEncoded is only being used for the non-Upload requests to Uplea, which even according to their API documentation is different than the Upload request. I admit the language on their API page suggests JSON encoding.

osfancy:
@Jaex I've committed the changes and also hopefully addressed the resx diff issue; VS wasn't even seeing that diff. Thanks for your patience.

ShareX.UploadersLib/Forms/UploadersConfigForm.resx
```diff
@@ -13483,4 +13483,415 @@ Using an encrypted library disables sharing.</value>
...
+  <data name="label4.Location" type="System.Drawing.Point, System.Drawing">
+    <value>17, 22</value>
+  </data>
+  <data name="label4.Size" type="System.Drawing.Size, System.Drawing">
+    <value>79, 13</value>
+  </data>
+  <data name="label4.TabIndex" type="System.Int32, mscorlib">
+    <value>14</value>
+  </data>
+  <data name="label4.Text" xml:space="preserve">
```
Jaex: You were changed control name but resx still using old name.

osfancy: Thanks for pointing that out @Jaex. That was actually the original 10 entries for the control left in by VS after the rename in designer. The new entries for lblUpleaEmailAddress were added to the same file, but I hadn't expected that it would not remove the references (or update) those for the old control name. I've just removed all references to label4 (the old control name) in the resx and checked it in. I even did a solution wide search just now to be doubly sure and the only references to label4 are in MovableShowColorForm.Designer.cs but that should be unrelated.

Jaex:
Thanks.

Jaex:
Is there specific reason you made api key textbox disabled? Because people could just copy it from http://uplea.com/api page instead of login from ShareX. Also now you can't logout (which means removing api key).

osfancy:
@Jaex Initially I was following the workflow implied by the get-my-api-key action, but realized the same thing that you just mentioned - which is that users could simply provide the API key in place of the username/password. I like the project and would like to have an opportunity to contribute further so consequently had planned to submit some further enhancements to that Uplea tab, including a logout option. Currently a user would need to fail logging in to blank out the API key - which I agree is not ideal.

Jaex:
Logout option will be misleading. Just need to make api key section editable so people clear it theirself.

Also there is no reason to store username and password in config. Because only api key is required.

osfancy:
Fair enough - will take that approach and add a link to the page with their API key.

Jaex:
Why are you adding link, people can already login inside ShareX.

osfancy:
The link would be there if we wanted them to only use an API key to connect w/o a username or password. If we want to have them still use their username and password to connect but simply not store the password (and/or username) after the the API key is retrieved then then that's simple enough too, and probably the better option and may have been what you were originally suggesting.

Jaex:
I meant that it could be optional way not only way. Also there is no reason to store username and password after login and received api key.

osfancy:
Agreed then - I think we're on the same page. Thanks for the clarification.

PR #1904, opened 2016-09-12
Update AmazonS3.cs

cursedtoast:
DreamObjects hostname changes in September. This line needs to be changed so functionality doesn't break.


Jaex:
Your link was containing invalid character `'` therefore I committed updated link.

cursedtoast:
Thanks for that. Probably an error in my copying the URL. Thanks for the quick approval

PR #1906, opened 2016-09-14
Updating Uplea Configuration

osfancy:
Update to not store Uplea username/password, button rename, and make API key input field editable.


ShareX.UploadersLib/Forms/UploadersConfigForm.resx
```diff
@@ -13652,7 +13652,7 @@ Using an encrypted library disables sharing.</value>
     <value>13</value>
   </data>
   <data name="btnUpleaLogin.Text" xml:space="preserve">
-    <value>Login</value>
+    <value>Get API Key</value>
```
osfancy: @Jaex Figured this button label would be more representative of the action being performed rather than the previous label.

Jaex:
I wonder why this pull request contains 13 commit when you were only did one commit for latest changes.

osfancy:
@Jaex I think it was because it included my previous commit history from the existing branch, probably an issue on my end.

Issue #1860, opened 2016-08-31
The invoked member is not supported in a dynamic assembly 


Permik:
How to reproduce:
1. Open main window
2. Go to Debug -> Debug log....
3. Press "Loaded assemblies"
4. Profit!

Tested on: Win 7 (64-bit) & Win 10 (64-bit)

The produced error:

```
System.NotSupportedException: The invoked member is not supported in a dynamic assembly.
   at System.Reflection.Emit.InternalAssemblyBuilder.get_Location()
   at ShareX.HelpersLib.DebugForm.btnLoadedAssemblies_Click(Object sender, EventArgs e)
   at System.Windows.Forms.Control.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnMouseUp(MouseEventArgs mevent)
   at System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.ButtonBase.WndProc(Message& m)
   at System.Windows.Forms.Button.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

ShareX 11.2.1 (Release)

```

Edit: After a while ShareX crashed and seemed to fix itself after restart.

Jaex:
This is weird I can't reproduce it when I click "Loaded assemblies".

TNTUP:
Happens to me too 
![ShareX](http://i.qcfb.ca/201609202065513.png)

Issue #706, opened 2015-05-24
Upload time with FTP is extremely slower


SirNickParks:
After updating to the latest version a day ago, I noticed a very large increase in upload time using FTP. Looked in the logs and found this new error?

    2015-05-23 21:24:14.093 - Unhandled exception:
    System.NotSupportedException: The invoked member is not supported in a dynamic assembly.
       at System.Reflection.Emit.InternalAssemblyBuilder.get_Location()
       at ShareX.HelpersLib.DebugForm.btnLoadedAssemblies_Click(Object sender, EventArgs e)
        at System.Windows.Forms.Button.OnMouseUp(MouseEventArgs mevent)
        at System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
        at System.Windows.Forms.Control.WndProc(Message& m)
        at System.Windows.Forms.ButtonBase.WndProc(Message& m)
        at System.Windows.Forms.Button.WndProc(Message& m)
        at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)

Here is the difference in times:
Before: 
2015-05-18 19:51:05.831 - Task completed. Filename: hRZWK.png, URL: http://nkp.us/hRZWK.png, Duration: 2527 ms

After: 
2015-05-23 21:24:15.176 - Task completed. Filename: LbiKJ.png, URL: http://nkp.us/LbiKJ.png, Duration: 13109 ms

Jaex:
Nothing changed in FTP codes for years and for months in FTP library so it is not possible for latest version to have different behavior than previous version. That error was to do with Loaded assemblies button in debug window.

PR #1913, opened 2016-09-18
Fixes the error message in debug screen when pressing loaded assemblies button, by not trying to include dynamic assemblies.

Permik:
Fixes the error message in debug screen when pressing loaded assemblies button, by not trying to include dynamic assemblies as these assemblies don't have a location.
Alternative way to fix this is referenced here: [https://github.com/dphoebus/Zbu.ModelsBuilder/commit/1cfb31f24ee6cf58c316519c8554384e1f22a0ec](https://github.com/dphoebus/Zbu.ModelsBuilder/commit/1cfb31f24ee6cf58c316519c8554384e1f22a0ec)
Fixes [706](https://github.com/ShareX/ShareX/issues/706) and [1860](https://github.com/ShareX/ShareX/issues/1860).

### DIFF
--- a/ShareX.HelpersLib/Forms/DebugForm.cs
+++ b/ShareX.HelpersLib/Forms/DebugForm.cs
@@ -25,6 +25,7 @@
 
 using System;
 using System.IO;
+using System.Linq;
 using System.Reflection;
 using System.Text;
 using System.Windows.Forms;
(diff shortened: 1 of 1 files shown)


Jaex:
I still wonder which dynamic assembly causing this. Because I can't reproduce it.

Jaex:
That was question by the way.

Permik:
I'll look into that today.

Permik:
Just to update:
This bug was a trickier one and I just can't seem to reproduce on my debug builds (yes, I'm building then without my fix)

PR #2005, opened 2016-10-16
Fix trim edge checking

jvitkauskas:
TrimTopEdge > 0 was checked twice instead of TrimRightEdge


Jaex:
Oh wow nobody reported all this time this issue :D

PR #2006, opened 2016-10-16
Use volatile variable for double locking

jvitkauskas:


Jaex:
I'm not sure why this needed for.

jvitkauskas:
Well, maybe you don't need that double locking at all? I have checked the code and it seems that you are initializing _taskbarList once. You can probably replace it with:
```cs
private static readonly ITaskbarList4 TaskbarList = (ITaskbarList4)new CTaskbarList();

static TaskbarManager()
{
  TaskbarList.HrInit();
}
```

jvitkauskas:
I have added aforementioned changes

PR #2007, opened 2016-10-16
Dispose MemoryStream objects after using them

jvitkauskas:
It is recommended to dispose MemoryStream objects after using them

### DIFF
--- a/ShareX.HelpersLib/Cryptographic/TranslatorHelper.cs
+++ b/ShareX.HelpersLib/Cryptographic/TranslatorHelper.cs
@@ -88,14 +88,15 @@ namespace ShareX.HelpersLib
         {
             binary = Regex.Replace(binary, @"[^01]", "");
 
-            MemoryStream stream = new MemoryStream();
-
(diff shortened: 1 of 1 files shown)


Jaex:
I already dispose every single one of them (if I don't miss) but this TranslatorHelper.cs was from my oldest project and I was just copy pasted it to ShareX, this is why it had bad code left in it I guess.

How you managed to find it too? Are you have helper Visual Studio addon which can find not disposed IDisposeable codes or something?

jvitkauskas:
You can use PVS-Studio for finding some bugs: http://www.viva64.com/en/pvs-studio-download/

Install it and go to menu: PVS-Studio -> Check -> Solution.

I think trial version has a limit on how much clicks you can do on the bugs it found, but you can always go to file and line manually.

Jaex:
http://www.viva64.com/en/inspections/

`We are friendly to open-source projects. We do our best to let their authors know about defects we find and grant them temporary free licenses for our tool if necessary.`

Maybe I should contact them and request free license. But not sure what they mean "temporary". Isn't then it becomes same as trial.

PR #2012, opened 2016-10-18
CLI for Edit Image

### LINKED ISSUE #1990: CLI for Edit Image
I've configured a CLI so that when I select a screenshot.png file, it should open up that selected file in the Image Editor.  I've tried:

"C:\Program Files\ShareX\ShareX.exe" "%1" -ImageEditor
"C:\Program Files\ShareX\ShareX.exe" -ImageEditor

But neither works.  They launch the editor but then launch the File->Open dialog window to select the image to edit... but I've already selected that image.

What am I doing wrong with the CLI?

ncmr:
Added capability to pass a file path as a parameter to the ImageEditor
CLICommand so it doesn't prompt an open file dialog if a valid image path
has been given in the CLI already and fixes #1990.

### DIFF
--- a/ShareX/TaskHelpers.cs
+++ b/ShareX/TaskHelpers.cs
@@ -605,7 +605,7 @@ namespace ShareX
 
         public static void AnnotateImage(string filePath, TaskSettings taskSettings = null)
         {
-            if (!string.IsNullOrEmpty(filePath))
+            if (!string.IsNullOrEmpty(filePath) && File.Exists(filePath))
             {
                 if (taskSettings == null) taskSettings = TaskSettings.GetDefaultTaskSettings();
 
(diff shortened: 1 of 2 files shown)


ShareX/TaskHelpers.cs
```diff
@@ -618,6 +618,10 @@ public static void AnnotateImage(string filePath, TaskSettings taskSettings = nu
                     AnnotateImageUsingGreenshot(null, filePath);
                 }
             }
+            else
+            {
+                AnnotateImage(taskSettings);
```
Jaex: This will make loop if dialog is canceled. Not to mention when someone right click to edit image that person won't like getting dialog to select random image.

Jaex:
Why your profile looks deleted? I needed to link it in change log.

PR #2124, opened 2016-12-02
Add NVENC support to FFmpeg options.

S-NA:
This makes it easier for those who own NVIDIA cards to use NVENC encoder that FFmpeg provides.

### DIFF
--- a/ShareX.ScreenCaptureLib/Enums.cs
+++ b/ShareX.ScreenCaptureLib/Enums.cs
@@ -89,6 +89,8 @@ namespace ShareX.ScreenCaptureLib
     {
         [Description("x264 (mp4)")]
         libx264,
+        [Description("x264-nvenc (mp4)")]
+        h264_nvenc,
         [Description("VP8 (webm)")]
         libvpx,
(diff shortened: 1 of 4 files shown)


Jaex:
1 year ago I needed to compile my own FFmpeg to include NVENC encoder but look like now https://ffmpeg.zeranoe.com/builds/ includes NVENC too. Because of this: https://github.com/FFmpeg/FFmpeg/commit/bc4137d4aa3a75bbbbc6b20c6710d07b50dab0a9

I was using `-c:v nvenc` when doing testing and you are using `-c:v h264_nvenc` now, is there any difference?

Jaex:
Now I tried recording with NVENC and CRF not affects its quality because quality becomes too low. Need to use `-b:v` instead to configure quality. I can implement this myself if you allow. Because I was already implemented this before and later removed it.

S-NA:
+1; sounds good. Thank you.

Jaex:
Finding NVENC preset list with descriptions was too difficult:



and now I'm not even sure which one is most suitable for screen recording.

PR #2131, opened 2016-12-08
Support for text uploading to pastie.org

plutofever:


plutofever:
Opening new PR to fix conflicts

PR #2165, opened 2016-12-20
Update ImageHelpers.cs

gaelbenoit:
Index Out Of Range when using Blur filter on a portrait picture
y maximum is 'y < h' so newColors array should be sized from h, not w


Jaex:
Thanks.

PR #2203, opened 2017-01-14
Fixed line endings for Hastebin Text upload

thijsa:
I have a modified Hastebin installation that has some more features, and I use \n to extract the line counting from the source file. However, when I upload a text file through ShareX, it uses \r\n as a line ending. So this is easily fixed by doing `text.Replace("\r\n", "\n")`.
Everything will still work normally on a default installation of Hastebin.

Example:
Before: https://paste.projectdev.org/utuhijurir
After: https://paste.projectdev.org/sofugumufa

### DIFF
--- a/ShareX.UploadersLib/TextUploaders/Hastebin.cs
+++ b/ShareX.UploadersLib/TextUploaders/Hastebin.cs
@@ -76,7 +76,7 @@ namespace ShareX.UploadersLib.TextUploaders
                     domain = "https://hastebin.com";
                 }
 
-                ur.Response = SendRequest(HttpMethod.POST, URLHelpers.CombineURL(domain, "documents"), text);
+                ur.Response = SendRequest(HttpMethod.POST, URLHelpers.CombineURL(domain, "documents"), text.Replace("\r\n", "\n"));
 
                 if (!string.IsNullOrEmpty(ur.Response))
                 {


Jaex:
So just because your modified Hastebin unable to check for both \r\n and \n when counting lines you modifying ShareX code to force \n ?! Seriously?

thijsa:
\n is used by the browser too. The API documentation of Hastebin is very old, and doesn't say that. 
When you do a post request through your normal browser, instead of the ShareX uploader, you use /n too, and not the windows file ending, \r\n.

thijsa:
@Jaex Okay, maybe I brought it a little bit wrong, that it looked like it was only needed for my own Hastebin installation. When you upload something through the official Hastebin site, it uses the `\n` line endings. When you upload something through the ShareX uploader, it uses the Windows `\r\n` line endings. It's not that I don't want to change the API that it accepts / converts the Windows line endings, but it would be nice if ShareX would use the same line endings as Hastebin itself. A better description of the pull request would be "Use the official line endings for the Hastebin uploader".

Jaex:
I checked hastebin request when saving text file and request contains `\r\n` according to Chrome dev tool.

thijsa:
Well that's weird, since it's stored as \n. Then there shouldn't be any reason why it shouldn't work.
Server-side storage: 
Server
I uploaded this through the website, not the API.

Jaex:
Well in any case if ShareX can upload `\r\n` and official hastebin implementation can show it without problem then your custom implementation must work with it without problem too. For compliance with default behavior.

thijsa:
Yes I know. I already fixed it in my own implementation. It was about the fact that ShareX uploads the text using the Windows line endings instead of the Unix line endings.

Jaex:
There is not a single text hosting web site in the world which have rule about using only Unix line endings.

PR #2229, opened 2017-01-26
Azure Storage support

wi5nia:
Added Azure Storage functionality as a File Uploader to the application.
There are 3 basic methods which are invoked:

- Check if the container exists. If not the container is created.
- Set the appropriate ACL on the container to enable public viewing of files.
- File uplaod to Azure Storage which return a URL


Jaex:
Thanks!

PR #2234, opened 2017-01-27
Pomf and Other Links

Outsanity:
These are the sites listed that do not work anymore when I went to them. it could be possible there's just some error on the server end or my end but do feel free to update them as you see fitting.

-nigger.cat seems to not be responding (possibly not in service anymore).
-pomf.hummingbird.moe links to some weird possible botting service.
-1339.cf links to random and annoying ads.
-comfy.moe seems to not be responding (possibly not in service anymore).
-files.plebeianparty.com seems to not be responding (possibly not in service anymore).
-maxfile.ro loads an empty page.
-minus.com has been shut down sometime since October 2016.
-pastee.org seems to not be in service anymore.
-SLiMG seems to not be responding (possibly not in service anymore).
-SomeImage seems to not be responding (possibly not in service anymore).


Jaex:
What is the point of removing them from readme when it is staying in software?

PR #2245, opened 2017-01-29
Update Pomf.cs

Ninn0x4F:
Switched from pomf.gocataclysm.com to pomf.space


Jaex:
Why there is conflict?

PR #2330, opened 2017-02-27
A better copy icon

Cereal-Killa:
A copy icon that doesn't look like a clipboard which is used for paste
instead. The icon is free for commercial use
(https://www.iconfinder.com/icons/173951/copy_icon).


Jaex:
All icons using same set there to be consistent with design yours not.

PR #2331, opened 2017-02-27
Master

Cereal-Killa:
A copy icon that looks like a copy icon, from the same icon set, Fugue.


Cereal-Killa:
From the same icon set, much more intuitive.

sharex vshost_2017-02-27_12-51-27

Jaex:
You know there is white copy icon in fugue set instead of blue one.

Cereal-Killa:
Didn't know, just found the white one here...

PR #2332, opened 2017-02-28
Added Table of Contents

Minimalistic:
Added a table of contents with link anchors to related content.


Jaex:
I was moved all readme contents to https://getsharex.com page recently, so readme page is outdated now and I'm gonna remove it because updating both place will be inconvenient.

PR #2355, opened 2017-03-10
Actions toolbar can now be placed on other monitors.

### LINKED ISSUE #2354: Actions toolbar is locked on one monitor
It is impossible to move Actions toolbar to any monitor other then the main one.

L1Q:
I used answers in this thread as a solution http://stackoverflow.com/questions/1317235/c-get-complete-desktop-size

Fixes #2354

### DIFF
--- a/ShareX/Forms/ActionsToolbarForm.cs
+++ b/ShareX/Forms/ActionsToolbarForm.cs
@@ -199,7 +199,7 @@ namespace ShareX
         private void CheckToolbarPosition()
         {
             Rectangle rectToolbar = Bounds;
-            Rectangle rectScreen = CaptureHelpers.GetScreenWorkingArea();
+            Rectangle rectScreen = SystemInformation.VirtualScreen;
             Point pos = rectToolbar.Location;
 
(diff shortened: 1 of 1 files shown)


Jaex:
Current way is not allows moving toolbar to top of taskbar because toolbar can stay behind taskbar even when it is top most window. You broken that so toolbar can disappear under taskbar now.

L1Q:
Is taskbar a special case? Are there other elements I should think of?

Jaex:
Method I used should be able to handle it in multi monitor. But because I don't have multi monitor not sure what is wrong.

L1Q:
For some reason "working area" is only one monitor.

Jaex:
Try using this: `return Screen.AllScreens.Select(x => x.WorkingArea).Combine();`

L1Q:
Cool! It worked!

Jaex:
Is it handle taskbar case too?

L1Q:
Yes, I can move it to other screen, but not over taskbar.

Issue #2365, opened 2017-03-13
File Downloader Behind "Transparent" Proxy - 401 Detection


ramseyjacob:
My company uses a "transparent" proxy which bounces a 401 back to any desktop program that tries to make an outbound request.  The File Downloader for "Capture > Screen Recording" tried to download the mpeg extension and failed with a 401.  I am able to make the download work by configuring the proxy manually through the Application Settings, but I had to hunt around.  It would be really nice if the following happened:

1. When the file downloader receives a 401, make the dialog more useful.  Right now I can't read the full message that is shown.
2. Check specifically for a 401 response code and offer a link directly to the Application Settings / Proxy screen so the user doesn't have to hunt around.
3. Cancel the screen capture when 401 detected - the user has to begin the process again after the proxy is configured, so they are forced to escape out of the capture mode.  It would be better just to cancel it for them if a 401 is detected.

**Steps to Reproduce**

1. Install fresh copy of ShareX on a network with a transparent proxy
2. Click "Capture > Screen Recording"
3. Authorize download of mpeg extension
4. Experience error: partial message shown, capture mode still initiated

I would love to contribute this change if deemed necessary.  I'm a .NET web developer so it might take me a bit to get up to speed on desktop dev.

PR #2367, opened 2017-03-14
Return false when the user cancels out or download fails.  Capture HT…

ramseyjacob:
…TP 401 response during file download and print a helpful message to configure proxy.

Addresses #2365


Jaex:
Unfortunately I can't test this because I can't reproduce proxy error. ShareX auto update using this window too so hopefully there is no problem about it.

ramseyjacob:
Very cool, thanks!  I tested it by substituting the URL to the EXE to any URL that returns a 401, since that is the behavior of a transparent proxy.

PR #2374, opened 2017-03-21
Add Standalone EDFile

Seenko:


Seenko:
By the way, I have no idea what happened with **ShareX.UploadersLib/Forms/UploadersConfigForm.resx**,

I compiled it locally and it worked, just need to solve that conflict.

PR #2375, opened 2017-03-21
Add translation for Traditional Chinese

alantsai:
Thanks for the great program

I have added new translation for Traditional Chinese which is mostly used in Taiwan or Hong Kong, hence the language code I used is zh-TW

Apart from the resx files, I have also added the language code in the enum and used the Taiwan flag found here https://www.iconfinder.com/icons/92372/taiwan_icon#size=16 for the dropdown language switch icon.
I have checked the usage licence should be ok, but if there is another place which you regular use for the language flag icon, let me know and I can change it.

If anything else is needed please let me know

Thanks again and good job.


Jaex:
Thanks for translation!

PR #2388, opened 2017-03-27
PVS-Studio: fixed two errors

IvanKishchenko:
## We have found and fixed two errors using PVS-Studio tool.

PVS-Studio is a static code analyzer for C, C++ and C#.

### Fixes:

1. _[V3084](https://www.viva64.com/en/w/V3084/) Anonymous function is used to unsubscribe from 'PropertyChanged' event. No handlers will be unsubscribed, as a separate delegate instance is created for each anonymous function declaration. Greenshot.ImageEditor AbstractFieldHolder.cs 105_
> Before my fixes 'RemoveField' method not unsubscribed handler from 'field.PropertyChanged' event, because delegate on subscribing and delegate on unsubscribing it's different objects with same body. I added a dictionary with keys and handlers for storing it's after subscribing and correct unsubscribing.

2. _[V3118](https://www.viva64.com/en/w/V3118/) Minutes component of TimeSpan is used, which does not represent full time interval. Possibly 'TotalMinutes' value was intended instead. ShareX.UploadersLib SendSpaceManager.cs 66_
> Using 'Minutes' property of TimeSpan not represent full time interval, therefore I changed it to 'TotalMinutes'.

### DIFF
--- a/Greenshot.ImageEditor/Drawing/Fields/AbstractFieldHolder.cs
+++ b/Greenshot.ImageEditor/Drawing/Fields/AbstractFieldHolder.cs
@@ -26,6 +26,7 @@ using System;
 using System.Collections.Generic;
 using System.Drawing;
 using System.Runtime.Serialization;
+using System.ComponentModel;
 
 namespace Greenshot.Drawing.Fields
 {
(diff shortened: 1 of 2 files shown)


Jaex:
We were requested open source project license for PVS Studio back then because your company website was telling that you support open source projects and open source projects can request free license. But your company were requested us to write article about PVS Studio in exchange of license and therefore we couldn't get license. Because not everyone have good english to write articles. I couldn't even send that request email myself therefore. Not to mention in that web site page you were not mentioned that writing article was necessary to get that free license. Otherwise we wouldn't bother sending email in first place.

So if you were caring about open source projects that much instead you could give license to us when we requested it and we could find these issues ourself, that way you didn't have to find and fix it yourself and send pull request.

IvanKishchenko:
We can offer you another type of free license: https://www.viva64.com/en/b/0457/

It doesn't presuppose any feedback from your side (articles, mentions and so on). 
Also, I recommend reviewing my edits, because there are really errors in your project that need to be fixed.

PR #2389, opened 2017-03-27
Updated gfycat to use newer oauth2 api

Dinnerbone:
The config tab doesn't have an icon for gfycat, because I can't find one to use.

Supports anonymous or authenticated uploads, and gives the option to either post gfycats public or private.

Also might theoretically stop that embarrassing issue with incorrect uploads, that some people have mentioned.


Jaex:
Thanks!

ObsidianMinor:
Amazing! This might help with [my project](https://github.com/ObsidianMinor/Gfycat.Net)

PR #2460, opened 2017-04-24
Fix S3 Uploads with random paths

hclewk:
Fixes an issue where random paths are being calculated twice - once on upload and once on url generation - causing the URL path to be different than the upload path.


Jaex:
Thanks.

deanpcmad:
When will this fix be live? 😃

Jaex:
https://getsharex.com/docs/dev-builds

deanpcmad:
Thanks will try that

Issue #2324, opened 2017-02-26
"You are attempting to upload a large file." dialog issues


YisroelAdler:
When trying to upload a large file (100MB by default) with ShareX it will ask if you are sure that you want to continue. 

But it is very annoying that the popup dialog pops up behind all open windows (see screen share below), and sometimes one doesn't realize until later when he opens the ShareX window and see it doesn't upload or until he closes all other windows.

I think it is very important that this dialog should appear on top of everything.

![Alt Text](http://i.imgur.com/FKKds6S.gif)

Jaex:
I even bring that window front in form shown event, how it managed to stay back.

Svetomech:
I had the same issue w/ winforms back when I still used them. Found no real solution, only a workaround of creating a timer which periodically sets `TopMost` property to `true`.

YisroelAdler:
For me, it does not show in front of any window I tried it with (Chrome, Windows Explorer, Virtualbox and more.)

Jaex:
That window stays top of everything (Chrome, Windows Explorer) in my pc. And I can't make it stay behind any window.

Jaex:
Your ShareX runs as admin I think. Not sure why but that causing this issue.

YisroelAdler:
Why do you think my ShareX runs as admin? From what I see (see Task Manager below) it does not. I also tried it now with run as admin and it was the same.

![taskmgr_2017-02-26_15-09-47](https://cloud.githubusercontent.com/assets/7387839/23343210/cbe64048-fc35-11e6-9454-824811293fca.png)

Jaex:
Because I ran ShareX as admin and finally I could reproduce your issue.

YisroelAdler:
But as you can see in my Task Manager screenshot I'm clearly not running it as admin.

The only way for me that it does pop up in front of all windows is if the ShareX window is open (as a window in the Task Bar and not merely in the Notification Area) and it is minimized to the Task Bar. And in that situation it comes in front even if run as admin.

PR #2491, opened 2017-05-05
Made sure that ForceActive will actually force the window to the front.

Farami:
This will fix Issue [2324](https://github.com/ShareX/ShareX/issues/2324).

To reproduce:
1. Start ShareX
2. Do not bring ShareX into focus, just leave it as it is.
2. Select a large file and right-click "Upload with ShareX".
The dialog window used to open behind everything else and will now open properly.

### DIFF
--- a/ShareX.HelpersLib/Extensions/Extensions.cs
+++ b/ShareX.HelpersLib/Extensions/Extensions.cs
@@ -357,6 +357,9 @@ namespace ShareX.HelpersLib
 
             form.BringToFront();
             form.Activate();
+            form.TopMost = true;
+            form.TopMost = false;
+            form.Focus();
         }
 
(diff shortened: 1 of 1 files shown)


Jaex:
This is very bad idea. That method used in most of forms including top most forms and you literally making top most forms not top most with this code.

Farami:
I don't see any issues with this. When forcing a dialog window to the point isn't it acceptable to show it before other TopMost windows? It will not force the dialog to stay above the TopMost windows, it will just push it to the top once.

Jaex:
How you won't see issue with this. `form.TopMost = false;` you making TopMost forms not TopMost.

Farami:
Oh I apologize. I have misunderstood your first comment. You are absolutely right. That is indeed an issue.

PR #2492, opened 2017-05-05
Made sure that ForceActive will actually force the window to the front.

Farami:
See PR [2491](https://github.com/ShareX/ShareX/pull/2491).

This time with a check whether the window may have been TopMost = true already.


Jaex:
I really don't think making every window top most momentarily is good idea. It can cause unintended problems.

Farami:
I agree that it is not the most ideal thing to do, however it was the only way I found that actually got the window to the front every time without issue.

PR #2597, opened 2017-06-30
Added NextCloud uploader & reverted NextCloud support for ownCloud uploader

aStonedPenguin:
While NextCloud is a fork of ownCloud is it taking its own direction with development and should have its own uploader instead of hacking support into the ownCloud uploader.


Jaex:
Isn't NextCloud and ownCloud uploader codes 99% same? How you are okey with such duplicated codes. We as developers trying to not duplicate codes as not possible. It is literally 2 lines of codes difference I think and you duplicated 194 lines of codes for it.

aStonedPenguin:
Because any future development to either of them will only be applicable for one of them. It is a little bothersome to have duplicate code at this current time but looking ahead to future platform changes for these uploaders and looking at the end users experience this is a better solution.

Jaex:
Better solution would be renaming "ownCloud" to "NextCloud / ownCloud" and in configuration page having radio box for people to select their choice. But even now both API is compatible with each other so there is no need for such radio box and just changing name would be enough. And this solution would require no duplicate codes at all.

PR #2653, opened 2017-07-26
Updated Italian Translation

pjammo:


Jaex:
You didn't need to translate some strings like uploader tab name because Imgur always become Imgur in all languages :P

PR #2713, opened 2017-08-18
Added a manifest to the setup

Faithfinder:
So that the setup asks for admin priviliges.


Jaex:
That setup not used by users. It is for AppVeyor.

Issue #2820, opened 2017-10-05
Region capture is broken when using touch screen


mghiasi75:
When I tap the region capture in the context menu of the tray icon the app correctly enters the region capture mode. but then trying to choose a region is not working correctly.

**Expected behavior:** After entering region capture mode, user should be able to put his finger somewhere on screen. It should be treated as starting point. Then user can move theirs finger to some other point (while still touching the screen), and the point that user puts theirs finger away should be the end point. The region between the two points will be captured.

**Observed behavior:** The app treats the previous place that finger was (which is where the context menu item of region capture was) as the starting point. So when user puts theirs finger on the screen, starting point is already set; and by moving finger, end point is moving with it. So every region that user captures begins with the place of context menu item of region capture.

Jaex:
Duplicate of #2673

mghiasi75:
Sorry for duplicate.

PR #2821, opened 2017-10-05
Fixed an issue with touch screens not being able to select region/draw/select drawing correctly

### LINKED ISSUE #2673: Left corner square wrong position when using edit + touch
Tablet with touch input:

1. region capture
2. draw something into the screen
3. select area/select
4. draw from left buttom a square

=> the first clicked x,y position is wrong and always randomly selected somewhere in the top corner of the screen

Please fix this or implement a fix for easy touch input. For example a x1y1 x2y2 selection mode to draw a square so you tap two times and it selectes the square of the two points.

mghiasi75:
This is a pull request to fix #2673 and #2820.

The issue is that `InputManager` doesn't have the currently touched point location at the moment of `BaseShape.OnCreating()` and `ShapeManager.GetIntersectShape()`.

Since mouse moves continuously, this is not a problem (or at least not noticable). However, since no mouse movement happens between previous MouseUp and next MouseDown event when using touch, this issue becomes important.

So, forcing `InputManager` to update on the beginning of `ShapeManager.StartRegionSelection()` fixes the issue.

(At first I wanted to do it inside `BaseShape.OnCreating()` but then I figured that selecting existing shapes also suffer from the same issue in touch screens. So putting it on the beginning of `StartRegionSelection()` seems more logical)

### DIFF
--- a/ShareX.ScreenCaptureLib/Shapes/ShapeManager.cs
+++ b/ShareX.ScreenCaptureLib/Shapes/ShapeManager.cs
@@ -643,6 +643,8 @@ namespace ShareX.ScreenCaptureLib
                 return;
             }
 
+            InputManager.Update(); //If it's a touch event we don't have the correct point yet, so refresh it now.
+
             BaseShape shape = GetIntersectShape();
 
             if (shape != null && shape.ShapeType == CurrentShapeType) // Select shape


Jaex:
Thanks!

Jaex:
I wonder is node drag works correctly now, because node hit test happens on frame update and your cursor update happens after it.

mghiasi75:
By resize node drag, you mean the dots around a shape which can be used to resize it?

Jaex:
These are resize nodes:

mghiasi75:
I just tried them, and they didn't work with touch, but I'm not sure if the problem is (just) a bug similar to this. Because those nodes are too small to be grabbed by finger accurately anyway.

I guess adding a touch-mode to the app would be nice, making the up toolbar and these nodes larger so they can be tapped/grabbed by fingers.

mghiasi75:
On an unrelated topic, do you have a timeline of when you send updates to the Store? Can't wait to have this fix on the Store version 😁

Jaex:
I don't have timeline.

mghiasi75:
Does this change available in the master branch? Because the test I've performed about the nodes was on the source I've built.

Jaex:
Yes.

mghiasi75:
Looking at the code, these nodes should work since there's an `InputManager.Update()` in the `Update()` function right before checking if mouse is on top of those nodes. However, it doesn't. (I tried to make the rectangle bigger, still doesn't work *most of the times*)

It seems that `IsMousePressed(MouseButtons.Left)` is not consistently working for touch events. I'll try to investigate further on that.

mghiasi75:
The previous comment (the part about `IsMousePressed` not working) was not true. I've confirmed that it works properly with touch.

The problem is, 24px is still too low for touch screen on my high-dpi (12", 2160x1440) tablet. I've tested various numbers, and 120px was smallest first size that was comfortable to resize on my tablet.

However, this size can't be fixed, as DPI of different devices and monitors differ. So I'm thinking 3x larger nodes for touch screens, but also relative to it's scaling. (So actually 3 times 'scaling factor', which in my tablet results in 3*1.75*24px)

Jaex:
Your fix was not affecting resize nodes, this must be the reason why it not works.

mghiasi75:
I don't think so.

The default `InputManager.Update()` (that was present even before my fix) is happening before the resize not check happens. (on [this line](https://github.com/ShareX/ShareX/blob/master/ShareX.ScreenCaptureLib/Forms/RegionCaptureForm.cs#L383))

Also, as I said increasing the size to 120px made it work comfortably on my tablet without any code changes.

Jaex:
When that input manager update your mouse position still on old position because only when you touch that position updates.

makedir:
Still useless. Make a two click mode to draw a rectangle. Tap two times on the screen and them it draws a rectangle of the area.

PR #2859, opened 2017-10-17
Add panning to image annotate

L1Q:
When annotating image use middle mouse buttnon to pan image and all shapes with it.

Also fixes crop not moving shapes.


Jaex:
Great job 👍

PR #2866, opened 2017-10-20
Add drag and drop from history window

### LINKED ISSUE #1550: Image history: allow drag and drop of the image files

campbeb:
Drag and drop of entries in history list view copies file to drop location.

Closes #1550


campbeb:
Will rework.

PR #2879, opened 2017-10-25
Updated Italian Translation

pjammo:


Jaex:
Why every single line showing as diff in your commit? You must be using different line ending than project.

pjammo:
No idea, I'm using ResXManager.

Jaex:
I think your Git client causing that new line issues.

pjammo:
I upload files through browser.

L1Q:
Italian translation files caused issues for multiple people now. I believe it's because the way you upload them (probably bypassing .gitattributtes settings).
If you keep uploading files your way we will be forced to download your files and reupload them ourselves and nobody likes to bother doing this.
And it also means you will not be credited for your work.
So best solution for everyody would be if you learned how to use git client.

pjammo:
There's still the same issue using GitHub Desktop.
It always worked fine uploading through browser anyway.
What am I supposed to do now?

L1Q:
Your last commit is great!
Compare [previous](https://github.com/ShareX/ShareX/pull/2879/commits/ebeeb2369cdf4fbda9ee5a1df560b5760bf8f1d2) one where every line starting from `<?xml version="1.0" encoding="utf-8"?>` was a diff with [last](https://github.com/ShareX/ShareX/pull/2879/commits/3e2e736e2a49e4f880723faf328f2b5079f89e76) where only actual changes are marked as diff.
So now we only need a way to (re?)upload your previous work using same method.

Jaex:
I think easiest way to solve this would be, making new fork and pull them to new folder. Then copy your it.IT.resx files to new forked folder. Then push those changes using GitHub desktop client. Then make new pull request with it. Because your current fork already have bad new line files committed, it can be difficult for you to revert them.

PR #2882, opened 2017-10-26
Added option to remember editor window state, position and size

Jaex:
This option saves window state, position and size when window closing and when user re open editor this window states will be applied.


JaixTim:
I can't seem to find this option in the current build (14.0.1) - am I missing something or has it been moved/removed? The image editor window is certainly NOT remembering it's position or size for me.

Jaex:
> I can't seem to find this option in the current build (14.0.1) - am I missing something or has it been moved/removed? The image editor window is certainly NOT remembering it's position or size for me.

PR #2951, opened 2017-11-27
Set BackColor of all TabPages to "Window" to fix rendering issues on certain themes

RedMser:
The TabControl of WinForms has a bug which prevents the background of its TabPages to redraw when the BackColor of it is set to Transparent (its default value) and certain Windows Themes are used (in my case, the [Tenebrous Crystal](http://www.mediafire.com/file/g3ddeb6vb14ycm6/Tenebrous+Theme+by+MarlonBarca4ever.rar) theme).

An example of this bug can be seen here: tabcontrolissues

Unless another fix than the one proposed exists, this bug would make many forms of ShareX unusable when using themes where this behavior occurs. As a side-effect, UseVisualStyleBackColor is set to false.

Users not affected by this bug will not see any visual differences whatsoever.


Jaex:
https://msdn.microsoft.com/en-us/library/ms160620(v=vs.110).aspx

```Changing the value of the BackColor property automatically sets the UseVisualStyleBackColor property to false. If you want the TabPage background to render using visual styles but you want the child controls to inherit a BackColor value that you specify, set the UseVisualStyleBackColor property after you set the BackColor property.```

You could set `UseVisualStyleBackColor` back to `true`.

RedMser:
@Jaex Thanks for your reply. This might clarify the reasons for changing both properties:

After some testing, I have found out that the bug seems to correlate with *only* the `UseVisualStyleBackColor` property being set to `true`. As stated in the docs, changing the `BackColor` also disables that property, which fixes the issue.

If Visual Styles are enabled, `BackColor.Transparent` is the default `BackColor`, telling the `TabControl` to inherit the color from the parent. Without Visual Styles, `Control` is used as the background color, which isn't the "correct" default per-se, as the background color of Forms is always `Window` by default.

I am unsure how `UseVisualStyleBackColor = false` and `BackColor.Transparent` interact together, and whether that would be a better choice than assuming that Forms use `Window` for their `BackColor` (which all do, by the way).

Finally, contrary to how the name of the property makes it seem like, `UseVisualStyleBackColor = false` does not change how `TabPages` are rendered on *any theme* I have tested this change with so far (including default, high-contrast, and custom).

Jaex:
So if `UseVisualStyleBackColor` remains true and you use `SystemColors.Window` as BackColor, then your issue not solves?

RedMser:
Nope, I have tried all combinations after posting: only changing `UseVisualStyleBackColor` to `false` will fix the issue. Unless you find anything that breaks with this change, though, would anything speak against it?

Jaex:
Even if I can't find anything breaks in my PC still I'm worrying it can break someones UI.

RedMser:
[This answer](https://stackoverflow.com/a/4265040/7475278) talks about the property, when true, making the background color slightly brighter than the regular form's background color (which is basically not visible at all, with most themes).

Besides that, the [source code](https://referencesource.microsoft.com/#System.Windows.Forms/winforms/Managed/System/WinForms/TabPage.cs,d46e10d3196b6f08) of `TabPage` shows that a background image will be drawn only when visual styles are enabled.

While there is very little information on this topic online, this seems to be about everything that changes with visual styles, at least on TabPages. Could this be rolled out as a beta change, to see if this can cause any trouble for users in other environments (older versions of Windows, maybe even with Wine/VirtualBox)?

Jaex:
In dark Windows themes your changes causing tab back color to look different than tab header, see this screenshot:



right one is your change, left one is how it was before.

RedMser:
I assume it can be fixed by using a different `SystemColors`. On the theme I tried, `Window` looked the same as the background color with Visual Styles, so I assume there's another color which will work in your case as well.

A lot of the `SystemColors` look identical on my theme, and the native method call doesn't tell where it gets its color from.

I don't have a theme that I can test this with, but it should be straightforward enough to find a suitable replacement color in any case.

PR #2966, opened 2017-12-01
Updated Korean Translations for ShareX

123jimin:
I discovered that there were too much untranslated text when I was using ShareX.

Sorry for not keeping up Korean translations frequently...


Jaex:
Thanks.

Issue #2933, opened 2017-11-18
Artefacts preview window


OMGasm:
Clicking on a gif in the preview window for the full size preview shows various artefacts, some of which stay in the small preview window until the frame gets updated.
![2017-11-18_15-59-09](https://user-images.githubusercontent.com/1008183/32977554-1ee82426-cc7b-11e7-8ed8-1bdfaa0857fa.gif)
![image](https://user-images.githubusercontent.com/1008183/32977602-3277966a-cc7c-11e7-92c4-cc41d63d93a2.png)

Jaex:
If two gif is previewed in two different picture box this happens. As expected solution should be cloning gif image so that fullscreen preview won't have any connection to other gif used in picture box. Yet even after clone it still it is connected to other gif image somehow. So there is bug in .NET Framework that clone method not clones it properly. Therefore I can't fix it.

DogeAmazed:
The Clone-method is only an soft-copy, so the underlying stream is the same.

PR #3011, opened 2017-12-16
Fixed issue #2933

DogeAmazed:
Fixed the issue of #2933
There were two options:
1. Deep clone the image.
2. Disable the preview in the main window.

I choosed the second option because I think it's performance-wise better than the first option. Also, as the dedicated preview window is an dialog, the user can't interact with the main window, so there is no need for a preview.


Jaex:
This solution is good enough 👍

PR #3021, opened 2017-12-19
Allow to disable setting public ACL on files uploaded via S3

sylveon:
This mostly relevant when using [Origin Access Identity](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-restricting-access-to-s3.html)


Jaex:
Thanks.

PR #3038, opened 2017-12-30
Add void.cat uploader

v0l:
Somebody messaged me on discord to say that the void upload option didnt work on ShareX.

I wasnt aware that it was added here and noticed that it would never work so i added a correct uploader


v0l:
Just noticed it doesn't work if you have a new config, also renamed the description because being "void" doesn't make much sense

Jaex:
We talked in Discord.

PR #3046, opened 2018-01-01
Use a SharedAssemblyInfo to minimize duplication

sylveon:


Jaex:
I don't think `[assembly: AssemblyDescription("Screen capture, file sharing and productivity tool")]` should be used for library files.

sylveon:
If you look at the removed lines, it was used in libraries too, so I left it as it was.

Jaex:
It is not?

sylveon:
Ah yes, my bad. I'll fix that

PR #3050, opened 2018-01-02
Updated Italian Translation

pjammo:


Jaex:
Sorry but in your all pull requests you are changing line endings of each file and making diffs show all lines changed. Just yesterday someone had trouble pull requesting because of your files was showing up in his git client modified because of auto line ending correction and he lost time because of it. So I can't accept your pull requests with wrong line endings anymore.

Issue #1590, opened 2016-05-30
Skype window screenshot name broken - "tm" non properly escaped


QuAzI:
Sometimes URLs opened from internal ftp doesn't work because not all chars escaped correctly

![404_not_found_-_vimperator_firefox_2016-05-30_12-45-40_7lgiluoa](https://cloud.githubusercontent.com/assets/1897605/15646237/4908e528-2665-11e6-824a-c5aa6caeb494.png)

Naming settings
![sharex_-_task_settings_sharex_2016-05-30_12-53-30_ovpj1gur](https://cloud.githubusercontent.com/assets/1897605/15646363/e440810e-2665-11e6-8564-13114e5508fa.png)

Uploading settings
![sharex_-_destination_settings_-_cusersyardocum_sharex_2016-05-30_12-54-06_imsxxnxm](https://cloud.githubusercontent.com/assets/1897605/15646381/f450cb9e-2665-11e6-9b57-ea6f90699bcf.png)

classyjakey:
strip unicode characters from text before naming, then?

QuAzI:
Then it is necessary to add a little salt. Correctly, Ramsay?
Really all non latin chars will be removed in this case. Include half of my files.

mkykadir:
I have tested this on Google Chrome (53) and Mozilla Firefox (49). It didn't make any problem for non Latin characters since browsers should handle them. I have taken a screen shot of a window with a title that include Trademark symbols, ™, ®. The resulting address bars are 

for Chrome:

![trademark_symbols_ _ type_tm_and_ _text_signs_on_k_chrome_2016-10-02_09-25-14](https://cloud.githubusercontent.com/assets/3192210/19019102/4835017a-8882-11e6-89b6-4601117e955c.png)

for Firefox:

![trademark_symbols_ _ type_tm_and_ _text_signs_on_k_firefox_2016-10-02_09-25-27](https://cloud.githubusercontent.com/assets/3192210/19019104/54d3b192-8882-11e6-9cc2-a5b65736c314.png)

I dont know how file names are handling in FTP servers exactly. This test server is created with FileZille Server on Windows on local. Maybe server handles these characters or browsers I dont know but it seems like **not a problem**. But a selection can be added to setting which may say "remove non latin characters" or like that. If you copy the address bar text to a text editor

™ will be represented as %E2%84%A2
® will be represented as %C2%AE

QuAzI:
[...]

PR #3052, opened 2018-01-03
Dont use canonically escaped url

v0l:
https://msdn.microsoft.com/en-us/library/system.uri.originalstring(v=vs.110).aspx

https://github.com/ShareX/ShareX/issues/1590

### DIFF
--- a/ShareX.UploadersLib/FileUploaders/FTPAccount.cs
+++ b/ShareX.UploadersLib/FileUploaders/FTPAccount.cs
@@ -244,7 +244,7 @@ namespace ShareX.UploadersLib
             }
 
             httpHomeUri.Scheme = BrowserProtocol.GetDescription();
-            return httpHomeUri.Uri.AbsoluteUri;
+            return httpHomeUri.Uri.OriginalString;
         }
 
         public string GetFtpPath(string filemame)


Jaex:
AbsoluteUri: `Skype%E2%84%A2_-_quazi_node_Skype%200tFSqEiv.png`
OriginalString: `Skype%E2%84%A2_-_quazi_node_Skype%200tFSqEiv.png`

You change makes no difference?

v0l:
sc

The `&lrm;` seems to get removed when using AbsoluteUri. 

So the Url returned isnt matching the filename

Jaex:
I tested it with file name `Skype™` which is what he used I think in that issue.

v0l:
No i think its from the `%t` he has on the file naming, it takes the name from window capture. 
The window title that gets captured includes this extra marker.

Jaex:
He gave this link: `http://storage.catalogloader.com/screenshots/yar/Skype%E2%84%A2_-_quazi_node_Skype_2016-10-03_10-53-21_0tFSqEiv.png`

Which converts to `Skype™`.

v0l:
Yep that's the link that comes from AbosulteUri but the actual filename includes the left-right marker, i tested in similar conditions as he has with SFTP.

Issue #1914, opened 2016-09-18
When opening the debug window, unfocusing from the window and trying to open a new debug window from tray hangs ShareX


Permik:
When opening the debug window, unfocusing from the window and trying to open a new debug window from tray hangs ShareX.
Nothing more to explain.

campbeb:
Here is the problem, but I'm not sure the solution:
- When Debug window is opened, it starts capturing all of the log messages and showing them
- When Debug window is opened from tray, it starts saving the configuration files
- While writing log messages to first Debug window about saving the configuration files, it tries to open the new Debug window
- Program freezes on this `lock (loggerLock)`
- The second Debug window is trying to open, but gets stuck waiting for the lock, which never becomes available b/c it was locked writing messages to first Debug window

This only happens when opening from tray, because it causes config files to be saved.

PR #3054, opened 2018-01-04
Refocus debug window when already open

v0l:
Only allow one instance of the DebugForm to exist

https://github.com/ShareX/ShareX/issues/1914


Jaex:
WTF

PR #3100, opened 2018-01-28
Update Lambda URLs

mstojcevich:
No longer have λ.pw -- replaced by lbda.net


Jaex:
If people already have previous URL in their config, this not gonna replace it?

mstojcevich:
I have removed the domain redirect, so the existing URLs should at least work until I can write a better fix to update the config files that use the old domain (xn--wxa, which is no longer up)

mstojcevich:
The old URL should now be replaced if it's in the config. I'll probably add the ability to set a custom URL in a future PR.

Jaex:
You may want to use `URLHelpers.Combine` for `Config.UploadURL + response.url`

PR #3108, opened 2018-01-30
Check limit before uploading to Streamable

JeremyShih:
Streamable limits video over 1GB and longer than 10 minutes for direct file uploads. And I hope to check if video is under limitation before uploading starts.

### DIFF
--- a/ShareX.UploadersLib/FileUploaders/Streamable.cs
+++ b/ShareX.UploadersLib/FileUploaders/Streamable.cs
@@ -87,8 +87,11 @@ namespace ShareX.UploadersLib.FileUploaders
                 headers = CreateAuthenticationHeader(Email, Password);
             }
             UploadResult result;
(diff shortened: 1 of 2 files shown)


Jaex:
```using (FileStream fs = stream as FileStream)```

How this gonna work for non files?

I understand checking file size but checking video duration there is not good idea and can yield wrong results depending on whats uploading.

JeremyShih:
non-files? for example?
Users could only select file with OpenFileDialog or select files or folders and then click the "upload with ShareX" in the context menu for upload, right?

Jaex:
It also supports MemoryStream for example for gif. So you can't assume it is always FileStream.

JeremyShih:
What about checking the limits only in video uploading?
I have tried uploading a video file and uploading after screen recording, in both case the 'stream' will be 'FileStream'.
The settings shown in pic below.

PR #3113, opened 2018-02-01
Add %rf to use a random line from a file

AtlasTheBot:
This adds `%rf{path}` to Task > Upload > File Naming, which reads the file at `path` and uses a random line from it for the filename (Similar to `%radjective` and `%ranimal`)

https://suplex.me/EasyHariyama.png
An example of it in action

If the file is not found/not able to be read/some other error, the preview will display the error message, but trying to upload will popup an error window
https://gfycat.com/BothZealousEstuarinecrocodile

### DIFF
--- a/ShareX.HelpersLib/Properties/Resources.fr.resx
+++ b/ShareX.HelpersLib/Properties/Resources.fr.resx
@@ -1024,4 +1024,7 @@ Voulez-vous la télécharger ?</value>
   <data name="LinearGradientMode_ForwardDiagonal" xml:space="preserve">
     <value>Diagonale en avant</value>
   </data>
+  <data name="ReplCodeMenuEntry_rf_Random_line_from_file" xml:space="preserve">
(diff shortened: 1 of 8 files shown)


ShareX.HelpersLib/Properties/Resources.fr.resx
```diff
@@ -1024,4 +1024,7 @@ Voulez-vous la télécharger ?</value>
   <data name="LinearGradientMode_ForwardDiagonal" xml:space="preserve">
     <value>Diagonale en avant</value>
   </data>
+  <data name="ReplCodeMenuEntry_rf_Random_line_from_file" xml:space="preserve">
+    <value>Line aléatoire d'un fichier. Utilisez {chemin} pour indiquer le fichier</value>
```
sylveon: *Ligne aléatoire

sylveon:
Is it possible to make it use the error notification, like this instead of a big popup:

PR #3139, opened 2018-02-12
Cleanup and simplify the hotkey selection menu Fixes #2122

### LINKED ISSUE #2122: Add clearer indication to the user in the hotkey settings window when hotkeys are disabled
I filed #2115 yesterday because I managed to disable hotkeys by accident and didn't realize it.  The only clue I had was the color of the boxes on the right side of the screen, which I thought to click on/mouse over, but they don't have any tooltips on them either.

I'd recommend adding tooltips to those boxes, plus some kind of notification bar at the top or bottom of the screen to let the user know that nothing they do in that screen means anything if they don't enable the hotkeys.

stuntguy3000:
Now featuring tooltips.

### DIFF
--- a/ShareX/Controls/HotkeySelectionControl.Designer.cs
+++ b/ShareX/Controls/HotkeySelectionControl.Designer.cs
@@ -28,19 +28,13 @@
         /// </summary>
         private void InitializeComponent()
         {
+            this.components = new System.ComponentModel.Container();
             System.ComponentModel.ComponentResourceManager resources = new System.ComponentModel.ComponentResourceManager(typeof(HotkeySelectionControl));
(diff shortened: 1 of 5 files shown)


Jaex:
I never used {  } in switch case in whole project so it becomes inconsistent.

stuntguy3000:
My Java background is clearly leaking over here. Let me look into what can and can't be done with switches and make the right adjustments.

stuntguy3000:
Corrected.

Jaex:
Sorry but now I check how your changes looks on UI and didn't like color of buttons like this:



Compared to how it is before:



I also asked in Discord server which one people prefers and like me also everyone preferred old one too.

Issue #3157, opened 2018-02-19
More settings for image watermarks


Bluscream:
![](https://i.imgur.com/4mbnYi4.png)

In addition to `Placement`, `Offset`, `AutoHide` and `ImageLocation` i'd like to request:

- `SizePolicy` with the possible values: `Initial`, `Fixed` and `Relative`
- `Size` which is
    - greyed out when `SizePolicy` is `Initial`
    - accepts numbers when `SizePolicy` is set to `Fixed`
    - accepts percentage when `SizePolicy` is set to `Relative`

Eranikid:
I would like to take this

Bluscream:
@Jaex when can i expect this to be in a compiled version? :)

Jaex:
https://getsharex.com/docs/dev-builds

PR #3158, opened 2018-02-20
Added Size and SizeMode properties to the image watermark effect

RedMser:
Implements the suggestion of #3157.

`SizeMode` has the following values:
- `AutoSize`: size is determined by the image's size, the `Size` property does nothing
- `Absolute`: size is determined by the `Size` property, in pixels
- `Percent`: size is determined by the `Size` property, as a relative percentage of the image's size

### DIFF
--- a/ShareX.ImageEffectsLib/Drawings/DrawImage.cs
+++ b/ShareX.ImageEffectsLib/Drawings/DrawImage.cs
@@ -28,6 +28,7 @@ using System.ComponentModel;
 using System.Drawing;
 using System.Drawing.Design;
 using System.IO;
+using System.Windows.Forms;
 
 namespace ShareX.ImageEffectsLib
 {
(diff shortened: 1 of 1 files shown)


Jaex:
Don't you think auto size is misleading when it is not resizing at all?

RedMser:
What is your proposed solution? You can't easily disable properties in the editor.
I thought about automatically changing the mode when you specify a size, and vice versa, but not sure how it'll work out with programatically setting the values.

Jaex:
This:

>  //TODO: Is a custom enum for different watermark image sizes a better idea?

RedMser:
Right, if it's only a matter of naming, I have created a custom enum `DrawImageSizeMode` with the values renamed to `DontResize`, `Absolute` and `Percentage` respectively. Should be clear now that the default value won't actually do anything with the `Size` property (but of course, alternatively the `Size` property could be populated when loading an image, and use `Absolute` scaling then...).

Jaex:
Relative mode also can be added. For example if you want watermark to be 20% size of image size.

Jaex:
Percentage can be misleading too because people can think it is percentage of inserted image instead of relative percentage to source image.

PR #3199, opened 2018-03-18
Cleanup code

bidolah:
- Remove null check when result of the expression is always true
- Remove empty constructors
- Remove unnecessary "string.Format"
- Replace "Where(...).Count();" by "Count(...) " 
- Replace "Where(...).FirstOrDefault()" by ""FirstOrDefault(...)."

### DIFF
--- a/ShareX.HelpersLib/Extensions/Extensions.cs
+++ b/ShareX.HelpersLib/Extensions/Extensions.cs
@@ -610,6 +610,7 @@ namespace ShareX.HelpersLib
 
             if (index > -1)
             {
+                // TODO: fix it. Assigned to itself
                 lb.Items[index] = lb.Items[index];
             }
         }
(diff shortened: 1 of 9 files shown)


ShareX.ScreenCaptureLib/Shapes/Drawing/LineDrawingShape.cs
```diff
@@ -51,11 +51,6 @@ private void AdjustPoints(int centerPointCount)
                 newPoints[newPoints.Length - 1] = Points[Points.Length - 1];
             }
 
-            for (int i = 0; i < newPoints.Length; i++)
```
sylveon: Pretty sure removing this changes behavior. Instead of having an array filled with Points, you get one filled with null except the first and last entries.

bidolah: image

newPoints is array of structs "System.Drawing.Point" https://referencesource.microsoft.com/#System.Drawing/commonui/System/Drawing/Point.cs,a041be61667d4c9a

Any element of an array can not be null. That expression (newPoints[i] == null) always return false. So, we can remove this cycle without changes in program logic.

Maybe we need to copy all the elements of the array points to new points?

sylveon:
Thanks, I didn't knew Point was a struct.

-- 
Envoyé de mon Gmail

PR #3205, opened 2018-03-19
Return download link if the file is not an image

mltnhm:
I noticed this _a bit_ too late, non-images do in fact produce strange previews


Jaex:
You sent it right after release :(

Jaex:
By the way you can use: `Helpers.IsImageFile(path)`

coderobe:
What about other media data? I'm not sure but i think the preview endpoint works for both audio and video as well.

Jaex:
I never used ownCloud so I have clue.

mltnhm:
I tried it with audio and it just returned a `[]`

coderobe:
Should probably check which files return the correct data (and whether there might be another endpoint that returns the raw data instead of generating a "preview" for some files) instead of creating these case-by-case rules

PR #3225, opened 2018-03-23
add call to GetReply after using OpenRead or OpenWrite according to FluentFTP documentation

omkelderman:
According to https://github.com/robinrodricks/FluentFTP#file-transfer you need to call `GetReply()` after completing the transfer from `OpenRead()` or `OpenWrite()`

Without this fix I get consistently the problems as described on the discord server (half the image is not there) for large(ish) images. With this fix I get consistently zero issues.

couple notes:
1. I pulled a chuck of code in `UploadData` to its own method, since it appeared twice and it needed the fix in both, so it was easier to just put it like this. However I suck at creating good names of things, so if you dont like the name of that method, gimmy a better one ;) xd
2. I dont know for sure what the `ftpResult` thing will contain, but from what I can tell its the ftp server telling the client if the transfer was ok, so I figured I use that info as well when returning
3. I added the `GetReply()` to `DownloadFile()` as well, I didnt test that, but I assume that'll work just fine since the exact same way works for uploading.

### DIFF
--- a/ShareX.UploadersLib/FileUploaders/FTP.cs
+++ b/ShareX.UploadersLib/FileUploaders/FTP.cs
@@ -231,10 +231,7 @@ namespace ShareX.UploadersLib.FileUploaders
             {
                 try
                 {
-                    using (Stream remoteStream = client.OpenWrite(remotePath))
-                    {
-                        return TransferData(localStream, remoteStream);
-                    }
+                    return UploadData2(localStream, remotePath);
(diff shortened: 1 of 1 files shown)


Jaex:
Thanks for fix. I'm unable to test it because I don't have FTP server. We should ask who had this problem to download [dev build](https://getsharex.com/docs/dev-builds) for testing.

PR #3125, opened 2018-02-05
Streamable upload limit

JeremyShih:

PR #3274, opened 2018-04-01
Add streamable upload limit (#1)

JeremyShih:
* Check limit before uploading to Streamable

Streamable limits video over 1GB and longer than 10 minutes for direct file uploads. And I hope to check if video is under limitation before uploading.

What kind of formats can I upload?
Basically anything that resembles a video file.
https://support.streamable.com/frequently-asked-questions

### DIFF
--- a/ShareX.UploadersLib/FileUploaders/Streamable.cs
+++ b/ShareX.UploadersLib/FileUploaders/Streamable.cs
@@ -26,7 +26,10 @@
 using Newtonsoft.Json;
 using ShareX.HelpersLib;
 using ShareX.UploadersLib.Properties;
+using Shell32;
+using System;
 using System.Collections.Specialized;
+using System.Diagnostics;
 using System.Drawing;
 using System.IO;
 using System.Threading;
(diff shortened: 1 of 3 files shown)


matthewburnett:
If this is a duplicate of #3125 you should close it.

keichinger:
This is the cleaned up version of #3125 (in terms of commits). The previous PR has already been closed.

sylveon:
- Don't implement IsVideoFile yourself, there already is `Helpers.IsVideoFile`
 - `as FileStream` is unsafe because you could get passed a `MemoryStream` and then your `fs` will be null, resulting in a NRE because you don't do any null checks (which also means `GetDuration` would need to be able to not have to read from disk for you to be able to remove the `as` cast)
 - You can use the `Length` property of `Stream` instead of doing `new FileInfo(fs.Name).Length`

JeremyShih:
@sylveon 
* I will use the method `Helpers.IsVideoFile` instead.
* The owner also told me about that. A video file will not reach `uploader.Upload` as a `MemoryStream`, so I put it inside `if (Helpers.IsVideoFile(fileName))`.

sylveon:
While this is true today, it might not in the future and break the Streamable uploader

ShareX.UploadersLib/FileUploaders/Streamable.cs
```diff
@@ -85,7 +87,29 @@ public override UploadResult Upload(Stream stream, string fileName)
...
+                GetDuration(fs.Name, out TimeSpan duration);
+                isOverDuration = duration > TimeSpan.FromMinutes(10); // video is over 10 minutes
+                isOverSize = new FileInfo(fs.Name).Length > 1024 * 1024 * 1024; // file is over 10GB
+                
+                if (isOverSize || isOverDuration)
+                {
+                    result = new UploadResult
+                    {
+                        IsSuccess = false,
+                        Response = "There is a 10 minute limit on video duration and a maximum file size of 1GB for direct file uploads."
```
stuntguy3000: Does this need localization?

ShareX.UploadersLib/FileUploaders/Streamable.cs
```diff
@@ -85,7 +87,29 @@ public override UploadResult Upload(Stream stream, string fileName)
...
+                    };
+                    Errors.Add("Video size or duration over Streamable limit(1GB and 10 minutes)");
```
stuntguy3000: Does this need localization?

ShareX.UploadersLib/FileUploaders/Streamable.cs
```diff
@@ -155,6 +179,25 @@ private void TranscodeFile(UploadResult result)
...
         }
+        private bool GetDuration(string filename, out TimeSpan duration)
+        {
+            try
+            {
+                var shl = new Shell();
+                var fldr = shl.NameSpace(Path.GetDirectoryName(filename));
+                var itm = fldr.ParseName(Path.GetFileName(filename));
+
+                // Index 27 is the video duration [This may not always be the case]
```
stuntguy3000: What happens if Index 27 is _not_ the case? Does everything break?

Issue #3080, opened 2018-01-16
The auto capture function of the tool..


c933103:
.... a few days ago, I was going to try the function of automatically make a capture of the screen after x seconds, which I thought I failed but actually it succeeded, which generated a few problems:
1. these auto-screenshots mixed together with my regular screenshots that I am not sure how to separate them
2. these auto-screenshots get automatically uploaded onto the cloud image hosting and I am not sure how to delete them
3. also I am not sure about how to delete them from ShareX's history
And that lead to a few more suggestions:
4. Would it be possible to have on screen visual indication that auto capture is ongoing at least have the option for such an indication to show up and especially when capture region is not full screen
5. Would it be possible to have an option for audio notification when each of these periodic capture task a done
6. Would it be possible to have an option in ShareX toolbar menu that tell users such capture is occuring and also provide a button to quickly stop such a capture?
7. Would it be possible to have a different set of post-capture action for these automatic tasks than regular uploads?

Jaex:
When auto capture running there is new tray icon, which shows it is running and allows you to stop it by clicking icon:

![](https://jaex.getsharex.com/2018/01/huDzu3bnfi.png)

vatterspun:
> these auto-screenshots get automatically uploaded onto the cloud image hosting and I am not sure how to delete them

I think you can right-click on the images, right-click, and select deletion URL.  You'll then have to visit each of these URLs, which admittedly might take some time.

![image](https://user-images.githubusercontent.com/1289134/35188980-e6d3bef2-fe06-11e7-8bac-825c583c243a.png)

> Would it be possible to have a different set of post-capture action for these automatic tasks than regular uploads?

I think it's possible to set that up via the hotkey creation view


> when recording a screen, there is a timer showing how much seconds have been recorded. Would it be possible to add a similar timer for autocapture showing how many seconds left for next capture?

Do you expect, if you do full screen auto-capture, that this timer get captured?

c933103:
>Use task setting to override after-capture setting and a stop button

I think it would be a good idea if these settings are integrated to the menu for the default auto capture task

>Do you expect, if you do full screen auto-capture, that this timer get captured?

It can optionally be shown and disappear when the timer reach zero

vatterspun:
> I think it would be a good idea if these settings are integrated to the menu for the default auto capture task

Can you suggest how you picture that getting implemented?

wolfen351:
I'd like to add my 2c here - this is a wishlist, feature request:

[...]

PR #3276, opened 2018-04-02
Respect Hotkey Overrides with Auto Capture

VodBox:
This change means that whether it's "Start auto capture using last region", or just opens the "Auto Capture" window, that any task settings overrides associated with the hotkey will be respected.

A use for this change is to be able to create screenshots for a timelapses while still using ShareX as normal. By overriding the capture path and after capture tasks, a user can have the auto captured images be in a separate directory and not be uploaded, without having to make those changes globally.

### DIFF
--- a/ShareX/Forms/AutoCaptureForm.cs
+++ b/ShareX/Forms/AutoCaptureForm.cs
@@ -51,6 +51,7 @@ namespace ShareX
         }
 
         public static bool IsRunning { get; private set; }
+        public TaskSettings TaskSettings { get; internal set; }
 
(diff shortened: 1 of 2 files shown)


VodBox:
This commit should also help with some of the issues mentioned in #3080, namely 1, 2 and 7.

Jaex:
`TaskSettings taskSettings = TaskSettings.GetDefaultTaskSettings();`

You should set this in `else`.

VodBox:
Updated.

Jaex:
Thanks.

PR #3292, opened 2018-04-08
Add Confirmation to Abort Recording - Fixes #3287

### LINKED ISSUE #3287: Alert question after selecting Abort button.
Please that would be really nice if application would ask you for confirmation after clicking Abort button.
I just lost an hour of recording because I missed Stop button.

stuntguy3000:
This PR adds a lovely confirmation dialog when a recording is cancelled.

Also, allows the user to toggle this function.
image

### DIFF
--- a/ShareX/Forms/ScreenRecordForm.cs
+++ b/ShareX/Forms/ScreenRecordForm.cs
@@ -226,7 +226,11 @@ namespace ShareX
         {
             if (e.Button == MouseButtons.Left)
             {
-                AbortRecording();
+                if (!Program.DefaultTaskSettings.CaptureSettings.AskConfirmationOnAbort ||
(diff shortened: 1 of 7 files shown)


Torubloo:
Remember to add a confirmation when stopping too because i misclicked and save the recording, also one for when starting a recording, opening sharex, and taking a screenshot and closing sharex , etc

anyway i don't think this is needed but hey you can toggle it at least i guess

stuntguy3000:
Keen to hear other people's thoughts on a confirmation for _stopping_

I'm against it on the basis stopping is usually an intentional and non destructive action, whereas abortion is destructive.

Torubloo:
https://github.com/ShareX/ShareX/issues/3287

stuntguy3000:
@Torubloo This PR was made *for* that issue, if you didn't catch that

Torubloo:
ik i was tagging that issue in this pr so people can read it

Scrxtchy:
I've pressed abort accidently about 20% of the time

PR #3327, opened 2018-04-21
Add Google Cloud Storage

### LINKED ISSUE #707: Support for Google Cloud Storage
There's support for S3 and even Google Drive, I'm surprised there isn't any Google cloud storage support, it would be great to have, thanks.

It appears to work similar to S3 and Google Drive.
API Stuff: https://cloud.google.com/storage/docs/overview
A custom URL option akin to the current S3 implementation should also be included.

matthewburnett:
Resolves #707
image


Jaex:
Good work 👍

PR #3351, opened 2018-04-29
Fix uploadSession error in OneDrive for Business

SupSuper:
Microsoft APIs are dumb


tannaalok:
we are getting 400 error again. can you please verify if its an issue with MS API ?

PR #3352, opened 2018-04-29
Fix Google Drive permissions

### LINKED ISSUE #3303: Interested in using the app in our schools, but permissions. Argh.
Why full access to Drive? We can't do that for FERPA reasons.

Why not:
https://www.googleapis.com/auth/drive.file | Per-file access to files created or opened by the app. File authorization is granted on a per-user basis and is revoked when the user deauthorizes the app.

Heck, I evaluate apps for my own use with the same question: are they requiring the least intrusive permissions, or the most?

I realize the app is capable of retrieving files it hasn't created and that this is a _feature_. But permitting the app to run with an inability to access remote files it did not create would satisfy our security needs.  It's possible to offer both for approval but make this one optional:

https://www.googleapis.com/auth/drive | Full, permissive scope to access all of a user's files, excluding the Application Data folder. Request this scope only when it is strictly necessary.

Any chance of this changing?

SupSuper:
Make ShareX permissions less intrusive. Should fix #3303

### DIFF
--- a/ShareX.UploadersLib/FileUploaders/GoogleDrive.cs
+++ b/ShareX.UploadersLib/FileUploaders/GoogleDrive.cs
@@ -80,7 +80,7 @@ namespace ShareX.UploadersLib.FileUploaders
         {
             GoogleAuth = new GoogleOAuth2(oauth, this)
             {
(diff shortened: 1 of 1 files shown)


chauffer:
Why was this closed?

PR #3397, opened 2018-05-14
 Ukrainian translation update

6c6c6:
Fixes and updates in Ukrainian localization


6c6c6:
Sorry, I don't know how to pull only my latest commit to main branch. It seems, that I just had to create a new mirror of the original repository.

Jaex:
You literally translated non existent stuffs because your repo was months old. Like this file `ShareX.HelpersLib/Automate/AutomateForm.uk.resx` is not even exists anymore in ShareX. Also you re added my removed files back. So you should have pull latest changes from ShareX repo first and then work on it. But easiest way would be for you to re fork it so you would have up to date data and work on it.

Issue #3144, opened 2018-02-14
Azure Storage uploader should have option to remove $root from URL when container name is $root


samwise800:
I am trying to configure ShareX to upload to my Azure Storage service, that has a single root container, called $root (the standard name for the root container).

The purpose for this is so that files can be accessed directly from
`http://<name>.blob.core.windows.net/<filename>`  rather than
`http://<name>.blob.core.windows.net/<containername>/<filename>`

I have set up ShareX to upload to the $root folder, and it works, however the result URL returned is of the form
`http://<name>.blob.core.windows.net/$root/<filename>`

Could there be a checkbox added to the options that will, when the container name is $root, remove the extra part from the URL so that the returned URL is instead of the form
`http://<name>.blob.core.windows.net/<filename>`

Thanks

Scrxtchy:
This could be implemented alongside #2845

Lego6245:
This is also a duplicate of #2710

Jaex:
Fixed by #3436

Jaex:
@samwise800 I was doing testing and noticed that when upload path is used (for example: 2018/06) with $root then it fails because first part of upload path becomes container. And if I use $root in request url (like this: $root/2018/06) then it also fails. So there is no way to have upload path when using $root?

Issue #2845, opened 2017-10-15
Azure Virtual Folders


Scrxtchy:
Really easy issue,
Azure blob storage allows storing within a virtual folder, and can be seen from the Portal within folders, as well as it's used within the blob url.

This can easily be added by adding the path to use to the url and canonicalizedResource variables

`string url = $"https://{AzureStorageAccountName}.{AzureStorageEnvironment}/{AzureStorageContainer}/{virtualFolder}/{fileName}";` as an example

No extra code is necessary for the uploader. This will just change the HTTP request for azure to add the path.

Ideally, this would be using a similar system to the Upload Path from AmazonS3

Scrxtchy:
🤖🔫👀

Lego6245:
In theory this is achievable by just adding additional stuff to the end of your container (it's not that smart). So if you wanted a virtual directory `bar` in container `foo`, you could set your container to `foo/bar`. However, explicitly allowing you to do custom directories like the upload path in S3 is a good enhancement.

Jaex:
Fixed by #3436

Issue #2710, opened 2017-08-17
Azure custom domain support and containers


altano:
These are my settings:
![image](https://user-images.githubusercontent.com/1009/29397134-192c79d6-82d1-11e7-9eb6-95ad6d7a5805.png)

My custom domain for Azure does NOT include the container, though, so the resulting URL which does include the container (/public) is wrong. E.g. I need ShareX to convert the URL:

`http://terriblefiles.blob.core.windows.net/public/XXX`

to

`https://files.terriblefish.com/XXX`

but instead I get

`http://files.terriblefish.com/public/XXX`

I see that the https issue was fixed but is it possible to specify that the container shouldn't be provided as well with the current settings?

Scrxtchy:
should a checkbox be implemented so that the container will not be appended to the returned url?

altano:
I would say that either the checkbox would be definitely be fine, or the most flexible option I can think of is to just let people type in whatever they want to create a custom URL, something like:

```
+----------------------------------------------------------------------------------+
|                                                                                  |
|   Account Name                                                                   |
|   +-------------------------------------------------------------------------+    |
|   | ...                                                                     |    |
|   +-------------------------------------------------------------------------+    |
|                                                                                  |
|   Access Key                                                                     |
|   +-------------------------------------------------------------------------+    |
|   | ...                                                                     |    |
|   +-------------------------------------------------------------------------+    |
|                                                                                  |
|   Container                                                                      |
|   +-------------------------------------------------------------------------+    |
|   | ...                                                                     |    |
|   +-------------------------------------------------------------------------+    |
|                                                                                  |
|   Environment                                                                    |
|   +-------------------------------------------------------------------------+    |
|   | ...                                                                     |    |
|   +-------------------------------------------------------------------------+    |
|                                                                                  |
|   Custom Domain                                                                  |
|   +-------------------------------------------------------------------------+    |
|   | https://files.terriblefish.com/{container}/{path-to-file}               |    |
|   +-------------------------------------------------------------------------+    |
|                                                                                  |
|       Supported Variables in Custom Domain:                                      |
|         * container                                                              |
|         * path-to-file                                                           |
|                                                                                  |
+----------------------------------------------------------------------------------+

```

Lego6245:
[...]

PR #3436, opened 2018-06-06
Azure Storage Updates

Lego6245:
Addresses #2710, #2845, and #3144 with a variety of changes.

1) Support masking container name when using a custom domain.
2) Return paths without $root in the URL.
3) Allow for custom upload paths like S3.


Jaex:
Thanks.

PR #3540, opened 2018-07-17
Fixed multiple capitalization errors which prevented building.

b4ux1t3:
When trying to build this project, I was running in to "Invalid Resx file" errors. Upon closer inspection, I noticed that there were a few resource files that were capitalized on disk, but were not in the Resource files that were supposed to point to them.

sharexbuilderrors

I know there is a setting that can ignore these errors and build without matching case, but I wanted to be sure that someone who wanted to start helping could quickly get up and running without editing their environment too much.

I have addressed all of the errors which are preventing a build out of the box.


sylveon:
The windows filesystem is case insensitive.

b4ux1t3:
@sylveon The Windows filesystem is not case insensitive. For instance, "ShareX" and "sharex" are two different files in the Windows filesystem. The Windows Command Prompt and PowerShell use a matching algorithm and a load order to determine what a user means when they type an ambiguous command or filename.

You'll notice that I posted a screenshot of the build errors. As I mentioned in my PR, I know these can be ignored, but now they don't have to be! :)

b4ux1t3:
I think the term you're looking for is case-preserving. However, I don't think that's the case in modern Windows. I could be wrong. 

In any case (ha!), that doesn't change the fact that these build errors do exist. Discussing whether or not Windows itself is case-sensitive is beside the point of: I have a fix for a problem that I ran in to while building ShareX.

Even if there weren't build errors, best practices would dictate that the code should accurately reflect files that it uses to build the application. If Windows ever changes how the filesystem works, or if someone adds new resource files that have the same letters but different capitalization and have that file actually be something different, this problem could manifest itself as an incorrect GUI. Not ideal, obviously.

sylveon:
Because the filesystem is case insensitive and will be for a while because
of backwards compatibility, it's impossible to have a file with the same
name but different case. Try it. In explorer create a file named test.txt
and try to make another one named Test.txt

b4ux1t3:
Huh. You're right. It occurs to me that I've never actually tried that.

The build errors still exist, and still replicate on a newly-installed Visual Studio instance. And, as I mentioned before, best practices dictate properly reflecting asset names in code. Given .NET is moving to cross-platform, where it may be possible to build this code on and for other platforms, migrating to cross-platform best practices is probably a good idea. _Especially_ since it will not break any existing code, since, as you pointed out, Windows itself doesn't seem to care about the case. It's an easy win, so to speak.

I point back to my previous comment: I ran into an issue while building ShareX, and have provided a fix.

PR #3563, opened 2018-07-30
Update LICENSE

IMJ355:
update year to current year


sylveon:
The dates on the license represent last update of the license text (and as such last copyright update)

The text is fine as it is, and is also what the official version of the GPL has.

PR #3582, opened 2018-08-11
Auto save OCR result

YueLinHo:
Save the OCR result to the same name txt file:
image

### DIFF
--- a/ShareX/TaskHelpers.cs
+++ b/ShareX/TaskHelpers.cs
@@ -1072,12 +1072,12 @@ namespace ShareX
             {
                 using (FileStream fs = new FileStream(filePath, FileMode.Open, FileAccess.Read, FileShare.Read))
                 {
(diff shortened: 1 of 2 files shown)


Jaex:
Thanks.

YueLinHo:
@Jaex Thank you so much.

bmidy:
Hi All,

First thanks to you for this great Tool !!
I d like to use this but I got no success.


My working day is like this.
I use a lots of screenshot, and later I d like to have the possibility to search in my Screenshot folders for some texts.

So I d like to have an Auto OCR of the Screenshots done, and like said above

to have the resulting OCR in the same folder, with same filename, date, just with a .txt etxension instead of .png

Accoring to this post it  seems to be OK, but I could not ??

Thanks

jeifour:
Hi, this seems really useful. However, I don't understand how to achieve this result. @YueLinHo could you give me a hint how this works?

PR #3619, opened 2018-09-10
Auto expiring of shared links for OwnCloud/NextCloud

jdon:
Added the ability to auto expire shared links after a certain amount of days.

Uses the expireDate paramaeter from OwnCloud/NextCloud OCS share API as described [here](https://doc.owncloud.org/server/10.0/developer_manual/core/ocs-share-api.html#create-a-new-share)


Jaex:
Why do you store this as string: `OwnCloudExpiryTime` if it is just number and gonna be cast to int anyway?
There is little typo here: `AutoEpxireTime`

jdon:
Good points! I have fixed the typo and updated the type to be an int.

Jaex:
Thanks.

Issue #3611, opened 2018-09-05
Drawing: Step - Make step number/letter configurable


Chefkeks:
Hey there,

occasionally doing screenshots with **Drawing: Step** and on every new screenshot the steps are set back to 1 (or A) and a continuous numbering from e.g.
Screenshot 1 - Step 1 + 2
Screenshot 2 - Step 3
Screenshot 3 - Step 4
Screenshot 4 - Step 5 -7 
is not really possible. Or am I mistaken here and there is an option already?

Therefore it would be great to have an enhancement to define the starting number/letter for _Drawing: Step_ via the menu, so you can start at e.g. 8 for screenshot 5.

Btw. currently doing this workaround: Adding step 1, step 2 and step 3 for example and move them all in the same position so step 3 is above step 1 and 2 and you only see step 3 on the screenshot, but a real solution via an option would be great!

Thanks in advance!
Michael

Jaex:
Currently as workaround I'm just moving initial steps outside of screenshot area.

Chefkeks:
That's an even better workaround 😁 👍

Jaex:
Added by @Fyers

Chefkeks:
So nice 😁 👍 
Thank you @Fyers

Taroslord:
Additional question:
Is there a way to add a second count starting from 1 inside the same shot?
I would like to count from 1,2,3 and restart a second counting starting at 1 at a different position.

So I could do this: 1,2,3,1,2,3,4,5 etc.

Currently I create another screenshot to start counting from 1 in the same image.

Devar:
> Additional question: Is there a way to add a second count starting from 1 inside the same shot?

Also wondering this, sometimes the ability to start a separate count set would be handy without having to save/commit an image and re-open to apply a new count set on top.

PR #3624, opened 2018-09-17
added StartingStepNumber option for AnnotationOptions #3611

Fyers:
Added a shape options for step counter, where the user can set the starting value as suggested by #3611

### DIFF
--- a/ShareX.ScreenCaptureLib/Properties/Resources.Designer.cs
+++ b/ShareX.ScreenCaptureLib/Properties/Resources.Designer.cs
@@ -1559,6 +1559,15 @@ namespace ShareX.ScreenCaptureLib.Properties {
                 return ResourceManager.GetString("ShapeManager_CreateToolbar_ShapeOptions", resourceCulture);
             }
         }
+		
+		/// <summary>
+        ///   Looks up a localized string similar Value of first step options.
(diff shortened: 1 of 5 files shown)


Jaex:
Are you not hiding that option when other tools selected?
Also I'm not sure why you setting this: `AnnotationOptions.StartingStepNumber = 1;`
I can't see that you set initial value of control too.

Fyers:
> Also I'm not sure why you setting this: `AnnotationOptions.StartingStepNumber = 1;`
I set `AnnotationOptions.StartingStepNumber = 1;` because else it would always start with the previous selected first value. So if I put 8 as the first value and save my image, the next image would also start at 8 because the value is never reseted.

Would it be neater to reset the initial value in the eg. ShapeEditor's `form_Shown` or in an other place?

Jaex:
If you were not planning to keep starting step number permanent then why you stored it in settings instead of storing it somewhere inside ShapeManager.

Jaex:
Thanks.

WiliTest:
I don't understand, is it possible to change the step? How, where? 
One intuitive way would be simply to either add a menu on the step button, or (maybe easier) to change the number when we click on it.

Issue #3242, opened 2018-03-25
[feature request] completely disable history list


greatwroth:
response to #3240
I would like to see an option to completely disable the history list even during an active session as it blocks easy viewing of currently configured hotkeys.

campbeb:
To clarify, you aren't looking to disable the history list because you don't want a listing of captures. But, you want to be able to always see the hotkey configuration. Is that correct?

greatwroth:
I don't care for having a list of captures as I don't feel a need for it personally. I was thinking along the lines of an option to toggle the list on or off. But yes, my main concern is being able to see the hotkey configuration at a glance.

PR #3630, opened 2018-09-22
added HistoryShow-Option

Fyers:
Implemented feature request from #3242 

sharex_2018-09-22_13-11-07


Jaex:
I don't feel like this is good idea, this is why I didn't add it :(

PR #3632, opened 2018-09-22
Added support for choosing Aboo.se as file upload destination (v2)

NoExitTV:
Added full support for choosing Aboo.se as file upload destination by using their public API.
Documentation for their API can be found here.

This works well for image/video/audio/general-file uploads since aboo will show a preview page for all files at https://aboo.se/ul/{upload-id}. Images/audio and video will be shown/playable directly at that web page so one does not have to download all files to view them.


NoExitTV:
I noticed that CodeFactor complains about the line: https://github.com/ShareX/ShareX/blob/9c3f9fe05801199b846a2265425e12ed8e62e1e8/ShareX.UploadersLib/FileUploaders/Aboo.cs#L51

But that was fixed in this commit: 
https://github.com/ShareX/ShareX/pull/3632/commits/3a1ba454114110105af79ff6959b2308e5cda9b9

Why does CodeFactor still complain about this?

NoExitTV:
Closing this pull request due to the above issue... Will open a new one soon from a new fork!

PR #3633, opened 2018-09-22
Added full support for choosing aboo.se as file upload destination

NoExitTV:
Added full support for choosing [Aboo.se](https://aboo.se) as file upload destination by using their public API.  
Documentation for their API can be found [here](https://api.aboo.se).

This works well for image/video/audio/general-file uploads since aboo will show a preview page for all files at https://aboo.se/ul/{upload-id}. Images/audio and video will be shown/playable directly at that web page so clients can preview files before downloading them.


NoExitTV:
Note that CodeFactor now complains that the FileUploaderService and the sealed FileUploader is located in the same file, Aboo.cs. After a through walt-through of the documentation at https://getsharex.com/docs/code-contribution I'm now confident that the CodeFactor plugin is wrong.

I've also minimized the added code as much as I think is possible.

ShareX.UploadersLib/FileUploaders/Aboo.cs
```diff
@@ -0,0 +1,130 @@
...
+            {
+                AbooUploadResponse response = JsonConvert.DeserializeObject<AbooUploadResponse>(result.Response);
+                if (response.status)
+                {
+                    result.URL = response.data.file.url.full;
+                    result.ShortenedURL = response.data.file.url.minimal;
+                }
+                else
+                {
+                    Errors.Add($"Message: {response.error.message}, type: {response.error.type}, code: {response.error.code}");
```
tinybarks: I think you should instead use "Aboo.se error, message: ", makes it clearer that the message comes from your uploader.

ShareX.UploadersLib/UploadersConfig.cs
```diff
@@ -401,6 +401,12 @@ public class UploadersConfig : SettingsBase<UploadersConfig>
 
         #endregion Google Cloud Storage
 
+        #region Aboo
+
+        public AccountType AbooAccountType = AccountType.Anonymous;
```
tinybarks: See below.

ShareX.UploadersLib/FileUploaders/Aboo.cs
```diff
@@ -0,0 +1,125 @@
...
+
+namespace ShareX.UploadersLib.FileUploaders
+{
+    public class AbooFileUploaderService : FileUploaderService
+    {
+        public override FileDestination EnumValue { get; } = FileDestination.Aboo;
+
+        public override bool CheckConfig(UploadersConfig config)
+        {
+            return config.AbooAccountType == AccountType.Anonymous;
```
tinybarks: Not sure why this field is necessary because this is not set anywhere and therefore serves no purpose, not to mention that the service does not appear to offer accounts (and that "full anonymity" is their goal).

ShareX.UploadersLib/FileUploaders/Aboo.cs
```diff
@@ -0,0 +1,125 @@
...
+        public const string AbooURL = "https://aboo.se";
+
+        private const string AbooAPIURL = "https://api.aboo.se";
+        private const string AbooAPIUploadURL = AbooAPIURL + "/upload";
+
+        public Aboo()
+        {
+        }
+
+        public override UploadResult Upload(Stream stream, string fileName)
```
tinybarks: Unfortunately, I'm not sure if there is enough logic here to warrant an Uploader.

PR #3636, opened 2018-09-30
Nuke DesktopBridgeHelper from orbit

sylveon:


Jaex:
Even I want to get rid of DesktopBridgeHelper a lot, I'm worrying a lot of people still could be using Fall Creators Update:



So for a change which not even gonna add any new feature, losing all potential or current users not feels right to me. Because people who still using Fall Creators Update won't be able to receive new update :(

sylveon:
Copying my response from Discord:
> the kind of people using the store tbh are also the kind of people keeping their machine updated. Plus, you can check how many users are using 17134 from the store dashboard to see how much people we can lose. It's also possible to keep the older package in the package list so people using <17134 will just get served an old version.

Jaex:
I can't check how many users are using 17134 from store dashboard because that info only shows for health section and there is so few errors with such info. Some people not gonna be happy about this change but I didn't wanted your efforts to go waste therefore merging it.

PR #3637, opened 2018-10-01
Implement Backblaze B2 uploader

tinybarks:
This adds support for Backblaze B2 in ShareX:

image

This code does not add any external dependencies. 

There are some minor changes to existing code:
- `Uploader.GetResponse()` now has an additional flag to enable it to return non-2xx responses. This does not affect existing code. 
- `MathHelpers.Random()` now uses a PRNG that returns high-quality random numbers. While not used in the B2 code, this changes makes randomly generated file names (like `5vnQUQ.png`) more unique, more collision resistant, and less guessable (`System.Random()` is seeded with the system time, I think via [`Environment.TickCount`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.tickcount?view=netframework-4.7.2)). Because B2 allows you to upload the same file to the same path as a new "revision", it is important to make sure that randomly named files will be sufficiently unique, even if several people were to upload at once.


tinybarks:
UI looks a little different now: 

image

PR #3640, opened 2018-10-01
 MathHelpers: add CryptoRandom()

tinybarks:
`MathHelpers.CryptoRandom()` uses a PRNG that returns high-quality random numbers. This PR makes randomly generated file names (like `5vnQUQ.png`) more unique, more collision resistant, and less guessable (`System.Random()` is seeded with the system time, I think via [`Environment.TickCount`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.tickcount?view=netframework-4.7.2)). 

Some uploader targets (e.g. B2) are able to upload the same file to the same path as a new "revision" or "version", so it is important to make sure that randomly named files will be sufficiently unique.


Jaex:
Thanks.

PR #3643, opened 2018-10-01
Auto size "After capture" list to show no scroll bar

### LINKED ISSUE #3519: Option to remove tasks from "After Capture" window
With the new addition of the "Scan QR code" task, the "After Capture" window now has a vertical scrollbar, making it annoying to choose the last task (please see picture below).

I suggest adding an option to remove tasks, that we do not use, from the "After Capture" window.
Making the window scale based on the amount of tasks could also be an option.

I apologize if there is already a solution to this, but I could not find one.

firefox_2018-07-10_23-21-14

Fyers:
Resolved #3519


Jaex:
Thats why I'm using [CodeMaid](http://www.codemaid.net/) addon which fixes those problems on file save.

PR #3649, opened 2018-10-01
Use current screen location for ImageViewer

Fyers:
On multiple monitors it was a problem, that the ImageViewer would always use the default monitor and not the current monitor.


Jaex:
It was using `Bounds = CaptureHelpers.GetScreenBounds();` so it is supposed to use your whole screen but I guess because of this `this.WindowState = FormWindowState.Maximized;` it was just using active screen? In your code you are also just changing location but not size but because of maximized it handles size I guess. Which is not good practice.

Maybe you should use `Bounds = GetActiveScreenBounds();` and then change window state to be not maximized?

Jaex:
I meant: `this.StartPosition = FormStartPosition.Manual;`

PR #3654, opened 2018-10-06
TextDrawingInputBox: Properly handle Enter key in CJK input

### LINKED ISSUE #3621: The screenshot text annotation function conflicts with the Chinese Pinyin input method.| 截图文字标注功能与中文拼音输入法冲突
When you take a screenshot, you can use the Chinese Pinyin input method, type Pinyin, and press enter to submit the text. This will result in direct submission of the entire screenshot, instead of just submitting Pinyin input method candidates, which is very bad.
***
截图时，标注文字弹窗界面，使用中文拼音输入法，键入拼音，按下回车键 提交文字过程中，会导致提直接交整个截图，而不是仅提交拼音输入法候选文字，这很糟糕。
***
image

tinybarks:
Fixes bug where entering text in the text annotation tool with a CJK IME and pressing <kbd>Enter</kbd> (to accept a suggestion) will commit the text instead of just accepting the suggestion.

Fixes #3621.

### DIFF
--- a/ShareX.ScreenCaptureLib/Forms/TextDrawingInputBox.cs
+++ b/ShareX.ScreenCaptureLib/Forms/TextDrawingInputBox.cs
@@ -38,6 +38,8 @@ namespace ShareX.ScreenCaptureLib
         public string InputText { get; private set; }
         public TextDrawingOptions Options { get; private set; }
 
+        private int _processKeyCount;
+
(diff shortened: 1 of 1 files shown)


tinybarks:
I've tested this and it works as expected with Google Japanese Input, Microsoft Japanese IME, and Microsoft Pinyin on Windows 10 1803, while English input continues to work.

PR #3657, opened 2018-10-09
Update Vietnamese vi-VN translation

thanhpd:
Update Vietnamese vi-VN translation to be on par with the current version


Jaex:
Thanks.

PR #3688, opened 2018-10-25
Improve localization support

L1Q:
- move strings to resx
- create tokens for missing language names in about form
- mark uploader names invariant


Jaex:
Thanks.

PR #3695, opened 2018-10-30
Fixes unexpected clipboard content viewer behavior 

### LINKED ISSUE #1769: Upload from clipboard with content viewer bug
Text (same for other types of content) shown in content viewer when uploading from clipboard can differ from actually uploaded stuff after pressing OK button.

How to reproduce:
1. Copy text 1
2. Use Upload > Upload from clipboard...
3. When content viewer appears with text 1, copy some different text
4. Press OK in content viewer (showing text 1) 
5. ShareX uploaded text 2

CookieMichal:
Fixes #1769.
Clipboard content displayed in ClipboardContentViewer is now cached with it's type. It allows clipboard to be freely used, while ClipboardContentViewer form is opened. When "OK" button is pressed, cached content is uploaded.

### DIFF
--- a/ShareX.HelpersLib/Enums.cs
+++ b/ShareX.HelpersLib/Enums.cs
@@ -36,6 +36,14 @@ namespace ShareX.HelpersLib
         URL
     }
 
+    public enum EClipboardContentType
+    {
+        Default,
+        Image,
+        Text,
+        Files
+    }
+
     public enum PNGBitDepth // Localized
     {
         Default,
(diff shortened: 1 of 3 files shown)


Jaex:
```ClipboardContent = img.Clone();```

If ClipboardContentViewer dialog is cancelled then this image should be disposed.

CookieMichal:
You are right, will fix later today.

Jaex:
Thanks.

PR #3728, opened 2018-11-11
Pull request template

trollmad3:
This repo has an issue template, but not a pull request template.


Jaex:
We don't have issue template.

Jaex:
Same reason as #3730
Just having "Added ... feature" is enough for me. I don't want people to write wall of text when I could easily understand what it is just by looking at changes.

PR #3730, opened 2018-11-11
Add issue template

trollmad3:
This issue template is better then the default one.


Jaex:
Sorry but I don't prefer these long issue templates otherwise I would have added it myself. I prefer issues to be brief as possible to make it easier to read. Not to mention it will also deter user from creating issue too.

PR #3821, opened 2018-12-12
Added Eraser tool to Image Editor

LiborB:
Added the ability to erase parts of an image (really just a white pen). Placed next to the crop button.
image


stuntguy3000:
Shouldn't it erase to transparent?

PR #3828, opened 2018-12-12
Added history settings option to open the edit window on double click in history and image history

### LINKED ISSUE #3571: Default action on clicking file in history
After making a snip it goes into your history, is there a way to have it so the default action when doubleclicking on the filename it goes to the editor instead of the viewer / browser?

LiborB:
Fixes #3571

image


Jaex:
I think it would be better if it had like this list of actions:



That way it can be extended when needed.

Issue #3868, opened 2019-01-08
Nextcloud "use preview link" checkbox creates direct download link


kwint:
When I check "use preview link (nextcloud only)" shareX generates links with /download on the end, making it a direct download link. When I uncheck that option the /download part is removed, which does result in a nice preview. Checking the ownCloud 8.1+ compatibility doesn't change anything. My server runs nextcloud 14.0.4 and sharex is version 12.3.1. 

Both cases generate links with /download:
![image](https://user-images.githubusercontent.com/7907280/50860991-b76eeb80-1397-11e9-851e-f266fcd0d7c7.png)
![image](https://user-images.githubusercontent.com/7907280/50861004-c0f85380-1397-11e9-8cb7-02e33e2adfac.png)
![image](https://user-images.githubusercontent.com/7907280/50861112-07e64900-1398-11e9-9fcb-cce328f9cec9.png)

If someone can confirm that this is not something on my site I might look into fixing this myself, altough I don't have any experince in C#

Jaex:
You need to check direct link I think.

kwint:
Nope checking "Direct link" still returns a /download link.

I was just reading the code and think I found the "bug".
When the preview checkbox is checked it either add /preview or /download to the end of the URL. The /preview only works for images, not for other files. So when a file is not an image it automatically adds /download. So unchecking preview generates normal URLs, without /download or /preview.

So the solution for me was to uncheck the preview link options to generate preview links for files other than images. Which isn't how it's supposed to be, I think.

https://github.com/ShareX/ShareX/blob/407c644f1f5665fa81e859824d15b9bab29f0af0/ShareX.UploadersLib/FileUploaders/OwnCloud.cs#L186

Jaex:
I have no clue about how those stuffs works, I never used NextCloud or OwnCloud before. Was blindly implemented it.

PR #3869, opened 2019-01-08
Don't generate direct download links when PreviewLink is checked

kwint:
See issue: https://github.com/ShareX/ShareX/issues/3868
Not adding either /preview or /download to URL generates links that show a preview of a file. /preview only works for images (showing literally only the image in the browser). This change makes sure the logic is right.

Note: The logic is right so it should work. But I'm not sure if the syntax is right.

### DIFF
--- a/ShareX.UploadersLib/FileUploaders/OwnCloud.cs
+++ b/ShareX.UploadersLib/FileUploaders/OwnCloud.cs
@@ -181,9 +181,9 @@ namespace ShareX.UploadersLib.FileUploaders
                     {
                         OwnCloudShareResponseData data = ((JObject)result.ocs.data).ToObject<OwnCloudShareResponseData>();
                         string link = data.url;
-                        if (PreviewLink)
+                        if (PreviewLink) && Helpers.IsImageFile(path)
                         {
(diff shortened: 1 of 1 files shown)


kwint:
I need some sleep. Will test this later

kwint:
I have no idea why it didn't build the first time but I guess I won't complain...

timvisee:
I don't understand why this would fail:
https://github.com/ShareX/ShareX/pull/3869/commits/eb0c4a1c131cd79e39126df2054bf7ba7d01d783#diff-7e56b6f655170710aff17bbb1557290bL184

Yet this succeeds:
https://github.com/ShareX/ShareX/pull/3869/commits/0c43a86e6e0211197e9fc418550ac11e4d9b390b#diff-7e56b6f655170710aff17bbb1557290bR184

Could someone explain, possibly @Jaex ?

Jaex:
Somehow AppVeyor still used old commit I think, because error is same in both commits.

timvisee:
@Jaex It's weird though, because AppVeyor shows the correct commit hashes we're being checked out in all of the builds. I'd be really interesting if it did check out source from a different ref than what is shown.

But maybe that's what was going on indeed. Have had different inconsistency errors before with AppVeyor.

Issue #3771, opened 2018-11-19
[Feature Request] Pushbullet - Pushing text and pictures


Somethingweirdhere:
Currently, you can use Pushbullet only as a File-Transfer uploader. But Pushbullet natively supports text and images too, as far as I know. Is there any reason that this is not supported in ShareX? If not, I could try to implement this. I don't have much experience with C#, but worked with the Pushbullet API.

PR #3871, opened 2019-01-09
Pushbullet upgrade

Somethingweirdhere:
Adds information about filetype to push. Pushbullet on android can then open the Image/File easily, and also shows a preview for images. Text can now be opened, but is currently sent as a textfile instead of actual text. Should we maybe change that too?

#3771


Jaex:
That would require creating text uploader for pushbullet which would be duplicate efforts.

Somethingweirdhere:
@Jaex makes sense, then I guess we will leave it as it is. Thanks for merging!

PR #3890, opened 2019-01-17
Google photos update for Picasa deprecation

matthewburnett:
~~Not~~ Ready for merge, but open for reference and discussion.

Known issues:
- Cannot upload to user created albums. - New behavior, cannot fix.
- ~~Sharing public images requires a different permission scope and the entire process is 4 requests which makes uploads substantially longer.~~ Fixed
- Direct links no longer supported - New behavior, cannot fix.


Daltz333:
Might want to squash some of those commits. Looks kinda bad for a public repository.

Jaex:
Thanks 👍

PR #3897, opened 2019-01-20
Remove broken websites from Chevereto websites list.

Abu3safeer:
I have commented the website that no longer working + added the the website status.
So please double check those websites, some of the has working domain, but that domain is linked to nothing, so it shows "DNS_PROBE_FINISHED_NXDOMAIN".
some of them redirect, some of them are for sale.

to be honest I wanted to add some good websites to the list, but I don't know how to get the api.


Jaex:
Thanks.

PR #3917, opened 2019-01-29
Added Teknik Services support w/ OAuth Authentication

uncled1023:
This will require a new Client Key/Secret combo added in the ApiKeys on release.  This can be created by creating an account at https://account.teknik.io/Register and then logging into the [Developer Settings](https://account.teknik.io/Settings/Developer) to create a client.


Jaex:
https://github.com/ShareX/CustomUploaders/issues/153

It was not possible to upload to Teknik before without account and creating account was requiring invite code, that's why it was took years for me to obtain Teknik account just to be able to test it. So this is changed now? Because it would be pointless to add uploader which have private account system.

Now when I tried to upload image from web site, upload was successful but I couldn't open image URL. It was giving "This site can’t be reached" error. But when I tried to open it with VPN it was opening fine. So it looks like my country (Turkey) blocked your web site or opposite.

By the way I noticed that you still didn't fix app name in this page: https://help.teknik.io/Tools which I was reported to you long years ago. How difficult it is to write app name correctly?

uncled1023:
As of v3.0, Invite Codes are no longer necessary.  As well, in this PR, Authentication is not needed for requests to be made.

I will look into the Tools page and update the app name.

uncled1023:
The name has been updated.  As well, I can provide the client credentials if you cannot access the site due to a block by your country.

uncled1023:
Certainly.  Where would you like me to put the main helper class?  Maybe in https://github.com/ShareX/ShareX/tree/master/ShareX.UploadersLib/Helpers or https://github.com/ShareX/ShareX/tree/master/ShareX.UploadersLib/OtherServices ?

Jaex:
Thanks.

zenden2k:
It seems that after 1 hour "access_token" expires, but ShareX does not refresh it. File is being uploaded, but not linked with account, therefore file will be deleted after 1 day. Teknik.io server does not report any errors in case of invalid access_token. You need to use "refresh_token".

PR #3944, opened 2019-02-09
Update Dropbox.cs

frostmouser:
Dropbox redirects from dropboxusercontent.com to dropbox.com.
This shortens the link.


Jaex:
It is not. It redirects from dropbox.com to dropboxusercontent.com.

frostmouser:
Yea, i meant the other way,
Brainfarted a little.
With this change you'll be able to post the short link which redirects to full.

Jaex:
Redirecting means there is two different requests. So making people do two requests to shorten domain length slightly is not ideal. It will delay the time that people will see image.

frostmouser:
ok, i get your point.

PR #3952, opened 2019-02-14
Performance/CPU enhancements on region capture forms and image editor

hungyiloo:
### Region Capture Form & Image Editor
Instead of invalidating the form repeatedly on every OnPaint (a high intensity infinite loop, which chews up an entire CPU core) this change throttles invalidate calls to a maximum of once every 16ms (~60fps) when there is user input, and once every 250ms (~4fps) when there is no user input and repainting is not necessary.

The Pause and Resume on lost/got focus events still work as they did before.

### Light & Transparent Region Capture**
By checking if the selected region has changed, we can avoid reprocessing/redrawing the screen, saving CPU usage too.

---

I realize these changes slightly affect the display of the dotted border, but I think the CPU reduction is worth it for users who use slow devices or rely on battery power. Let me know if you think any amendments are needed.

### DIFF
--- a/ShareX.ScreenCaptureLib/Forms/RegionCaptureForm.cs
+++ b/ShareX.ScreenCaptureLib/Forms/RegionCaptureForm.cs
@@ -131,7 +131,7 @@ namespace ShareX.ScreenCaptureLib
             timerFPS = new Stopwatch();
             regionAnimation = new RectangleAnimation()
             {
-                Duration = TimeSpan.FromMilliseconds(1000)
+                Duration = TimeSpan.FromMilliseconds(200)
             };
 
             if (IsEditorMode && Options.ShowEditorPanTip)
--- a/ShareX.ScreenCaptureLib/Forms/RegionCaptureForm.cs
+++ b/ShareX.ScreenCaptureLib/Forms/RegionCaptureForm.cs
@@ -100,11 +100,17 @@ namespace ShareX.ScreenCaptureLib
         private Font infoFont, infoFontMedium, infoFontBig;
         private Stopwatch timerStart, timerFPS;
         private int frameCount;
-        private bool pause, isKeyAllowed, forceClose;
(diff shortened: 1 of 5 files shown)


Jaex:
You literally reverted all my changes to how it was before. Long years ago I was also using timers and it was giving choppy experience. Main problem with using timer is, it is not high resolution timer so it's interval can trigger with different timings like even if you set it to 15ms it can trigger 30ms later. So my aim was making it smooth as possible. Also not having high resolution timer is the reason why I couldn't add max fps option to limit it to for example max 100 fps with Thread.Sleep etc. too.

Also there is no such thing as rendering it only when necessary because there can be a lot of animated stuffs in screen which requires rendering all the time, for example ant selection lines which always moving. Sure you can calculate regions of all animated objects and only re draw that regions in screen but it can get very complicated when there is a lot of animated objects. But there is one optimization I could do which is when region capture/image editor window lose focus then it stops rendering. So if someone do alt tab to work on something else then image editor won't try to render for no reason.

Even I used your system before and know how choppy it can be, still gave it a shot and pulled your PR for testing and I was right. Magnifier following cursor is so choppy it feels like 10 fps animation with that random jumping etc. So it is not acceptable like this and if it were I would have keep it like this before.

For people who prefer lower cpu and battery usage, can use Region capture (Light) which also using timers like yours: https://github.com/ShareX/ShareX/blob/master/ShareX.ScreenCaptureLib/Forms/RegionCaptureLightForm.cs#L73

Or they can use proper image editor software which will render only when there is changes. But this is just region capture form modified to act like image editor with a lot of workarounds. If I wanted to make normal image editor instead of on screen annotation, I would have made rendering system like that too.

hungyiloo:
Thanks for considering my request and apologies if I caused any offence. I understand your decision. Keep doing the great work you have been doing so far.

PR #3957, opened 2019-02-18
Update pomf.cs

3k2:
added new hosts and removed dead


lucario:
All future pomf uploaders should be added to https://github.com/ShareX/CustomUploaders
Constant rewrite of embeded short life uploaders is a maintenance hassle.

3k2:
> All future pomf uploaders should be added to https://github.com/ShareX/CustomUploaders
> Constant rewrite of embeded short life uploaders is a maintenance hassle.

shouldn't then Chevereto to be also added to customUploaders section since there seems to be hassle too?

lucario:
Yes, but we are currently undergoing updates to the custom uploader format which will be released in 12.4 which will require old ones to be rewritten (existing will be automatically rewritten). Depending on how that goes, I may see to removing all embedded uploaders entirely and just keep a way to manually input domains, or remove the uploaders entirely and rely on custom uploaders repo.

PR #4025, opened 2019-03-14
Add support for Intel QuickSync

sylveon:
Intel QuickSync provides hardware H.264 encoding on Intel Sandy Bridge CPUs and up and hardware H.265 encoding on Skylake CPUs and up via a dedicated die on the processor chip, being faster than CPU encoding and more power efficient than GPU encoding


Jaex:
Thanks.

PR #4030, opened 2019-03-16
Add Mega Folder Select Form

Sayayaya:
I made a folder browser because I had problems using MEGA - The combobox could not display the full file paths and was so long that scrolling to an item was difficult and knowing what was selected impossible.

An image of the browser:-

https://mega.nz/#!s89iDICI!TKPMz6qReSoeHip26zsNUNb8UaJRtiKr3sCfPgxZgYo

The folder icon is opensource. Comes from https://feathericons.com/


Jaex:
Your code conventions all over the place. Whole project using upper case for functions yet you used both lowercase and uppercase randomly for functions:



Also randomly starting private variables with _ character:



Whats up with this double empty lines and spaces:







You didn't even bother naming controls or using same naming format we used for form controls:



You could use mega icon stored in resources for form icon instead of storing icon as base64 inside resx file.

For folder icons you could have used same icon set we used for all icons already: http://p.yusukekamiyamane.com/icons/downloads/fugue-icons-3.5.6.zip

Sayayaya:
I'm leaving all of that to you, I just wanted a working file uploader and I got one. Only sharing this back incase you want to do something with it.

Issue #3769, opened 2018-11-19
Default "Select tool" for Image Editor


Danielku15:
**Request**
Please note that this feature request might be a duplicate of #3044 which got stale and was closed by the bot. I hope by this more comprehensive feature request we might get the feature described here. 

Please add a new "select tool" to the image editor which allows users to select/edit/delete any shape that was added via the other drawing tools. It would be great if via an additional option users can decide whether the select tool is automatically activated after any other shape was drawn. Clicking again on the existing tool button in the toolbar should activate again the "select tool". 

The "select tool" should work similar as any other tool for the individual types. e.g. when the rectangle tool is active, you can hover the shapes and they get highlighted as selectable. When selecting you see the additional controls and potentially a double-click opens a editor (text). Once selected users can use the shown controls to edit the shape or delete them via keyboard (DEL). The select tool should allow this for all shape types on the screen without the need of switching to the individual tool. Clicking anywhere, where no shape is results in no action. When a user presses ESC while a drawing tool is active, it should activate the "select tool" to cancel the drawing operation. If ESC is pressed while the "select tool" is active, the existing close confirmation should be shown. 

**Background** 
I just moved from Greenshot to ShareX and so far it looks really nice. I have setup as default task that any screenshot is opened in the image editor for further annotation and decision whether to save or only copy it. 

One main thing I dislike on the image editor is that there is always a "drawing tool" active. For me this results often in some accidentally created shapes. In Greenshot I liked the behavior that the default tool was some sort of "select tool" where you could select and modify any existing shape in the image. Also after drawing any shape, it jumps back to the select tool. 

The behavior in ShareX is a bit tricky to use. If you want to edit or delete any shape, you need to select the correct option in the toolbar and then you can start modifying shapes of this type. Also it seems that freehand drawings cannot be selected and deleted after creation. 

**Dev Info**
It seems that the [ShapeManager class](https://github.com/ShareX/ShareX/blob/9d923aeb13d72d5bba9eec195c542f631baad96e/ShareX.ScreenCaptureLib/Shapes/ShapeManager.cs) is the area that will need mainly the adjustments. It hooks up all events and holds the tool information. The `StartRegionSelection` method has the logic in place that you can only select shapes of the current tool. There we could add the check that it needs to be the current tool or the select tool for selection. Then only the enum for the ShapeType, the UI behavior and the handling of the `ShapeType.Select` needs to be added. Maybe I'll give it a shot to add the feature and file a pull request.

[...]

PR #4058, opened 2019-04-04
Added select tool to image editor 

Danielku15:
This pull request adds a new select tool to the image editor as proposed in #3769 

Tasks: 
- [x] Add select tool
- [x] Change escape key behavior
- [x] Update select tool icon
- [x] Add new setting that will allows the user to switch between the select modes: "Keep select tool active after shape selection", "Select shape tool after shape selection".
- [x] Add new setting that will allows the user to switch between the draw modes: "Keep current draw tool active after shape drawing", "Switch to select tool after shape creation"
- [x] Fix Crop Tool and use correct icon


ShareX.ScreenCaptureLib/Shapes/ShapeManager.cs
```diff
@@ -90,7 +93,10 @@ private set
                     ClearTools();
                 }
 
-                DeselectCurrentShape();
+                if(previousTool != ShapeType.ToolSelect && currentTool != ShapeType.ToolSelect)
```
Jaex: Why there is no space after `if` here?

Jaex:
I wonder is it good idea to use rectangle selection icon like this in Photoshop:



when ShareX not supports selecting multiple shapes with rectangle selection.

And in that case also shortcut of <kbd>M</kbd> becomes invalid too.

Jaex:
In image editor mode when "Switch to selection too after shape drawing" is selected, crop tool not works.

Danielku15:
> I wonder is it good idea to use rectangle selection icon like this in Photoshop:
> 
> 
> 
> when ShareX not supports selecting multiple shapes with rectangle selection.

Any icon you would prefer? Initially I had the cursor icon in mind, but this is already used for drawing a cursor. Beside that also the icons image-map and layer-select would fit quite good. I like the move tool icons from paint.net but I could not find some icon like this in the fugue icon set: 
image

> And in that case also shortcut of M becomes wrong too.

Depends on how you look at it 😉 . In Photoshop the <kbd>M</kbd> stands for "Marquee" while in Paint.net it is rather "Move". <kbd>S</kbd> is already used for "Speech-Bubble" and can therefore not be used for "Select" anymore. Any proposal that you have as alternative? 

> In image editor mode when "Switch to selection too after shape drawing" is selected, crop tool not works.
> Also selecting "select tool" after crop tool drawing have issue too.

I will have a look on that. Likely I will limit this "feature/option" to the actual drawing tools. Then it should not impact the crop tool anymore. But first I will check how it actually affects it. A fix will come soon.

Jaex:
Using cursor icon for select tool and finding another icon for cursor drawing tool gonna be more easier I guess.

It is fine then if "M" means move.

Jaex:
stamp-cursor
I made this icon for cursor tool now. So current cursor icon can be used for select tool.

Danielku15:
I made some more adjustments and now the crop tool is working fine again. Also after cropping the setting is considered and the "select tool" is activated. Let me know if you discover more

ShareX.ScreenCaptureLib/RegionCaptureOptions.cs
```diff
@@ -73,11 +73,14 @@ public class RegionCaptureOptions
         public bool MenuCollapsed = false;
         public Point MenuPosition = Point.Empty;
 
+        public bool SwitchToDrawingToolAfterSelection = true;
```
Jaex: I think making this option `false` by default would make more sense for users.

Jaex:
Great work, thanks for implementing it!

Jaex:
I noticed now this issue I reported:

> Also selecting "select tool" after crop tool drawing have issue too.

Is still exists.

Danielku15:
> Is still exists.

I made various tests and it worked fine for me? Can I assume, as you merged the PR, that the issue disappeared? 

Also there are some other remarks. Will you take care of this decision or do you want me to follow up on anything? 😉

Jaex:
I just forget to test it again before merge :D I assumed you fixed it.

This is the issue:



Yea I was asking your opinion about should that option be false by default.

Jaex:
When you select shape with select tool and then change to any drawing tool then selection still stays active:



But if you change from one drawing tool to another tool then as intended it unselects previous shape.

Danielku15:
Okay, thanks for the info. I will try to fix the remaining 3 points and will file another PR:
- [ ] [Bug] Crop-Tool Selection must be cleared when switching to the select tool 
- [ ] [Bug] Select-Tool Selection must be cleared when switching to any other tool 
- [ ] [Change] Make the auto-selection of tools disabled by default

Jaex:
Found another issue which happens if you select resizeable shape first then not resizeable:

Danielku15:
Wow, looks like there are quite a lot of individual tool checks scattered across the code that rely that the current tool matches the shape. I will have a closer look if I can get this cleaned.

Jaex:
Hi @Danielku15, this release a lot of people were reporting that their annotation shapes border size is resetting to 0 and their shapes becomes invisible. Recently I noticed in here #4418 that issue is actually related to "Select tool".

Reproduce steps:

1. Select any text tool and set border size to 0 then add text to screen.
2. Use "Select tool" and select the text.
3. Switch to "Line tool".

You will see that line tool border size changed to 0 which is previously selected shapes border size with cursor tool. When you have time can you check this issue?

Jaex:
I fixed the issue now.

PR #4068, opened 2019-04-08
Fixing Russian translation

maximmax42:
Current version of translation, albeit more compact, makes it impossible to distinguish between file, image and text uploaders in the list of uploaders in Task Settings -> Upload -> Uploader Filters. This pull fixes that.


Jaex:
Thanks.

PR #4083, opened 2019-04-15
Fix issues for select tool 

Danielku15:
This is the follow up PR of #4058 where the select tool was added. This PR is supposed to fix: 

- [ ] [Bug] Crop-Tool Selection must be cleared when switching to the select tool 
- [ ] [Bug] Select-Tool Selection must be cleared when switching to any other tool 
- [ ] [Change] Make the auto-selection of tools disabled by default
- [ ] [Bug] Improper behavior regarding resizable and not resizable shapes.


Danielku15:
Wrongly created PR -> Closing.

PR #4084, opened 2019-04-15
Fix issues for select tool

Danielku15:
This is the follow up PR of #4058 where the select tool was added. This PR is supposed to fix: 

- [x] [Bug] Crop-Tool Selection must be cleared when switching to the select tool 
- [x] [Bug] Select-Tool Selection must be cleared when switching to any other tool 
- [x] [Bug] Improper behavior regarding resizable and not resizable shapes. 
- [x] [Change] Make the auto-selection of tools disabled by default
- [x] [Bug] Auto-Switch to Select tool after drawing: Shape should stay selected after drawing. 

**Tested Scenarios:**

1.  _with switch to select tool option active_: Fully use crop tool -> Cropping not cancelled while drawing region 
2. _with switch to select tool option active_: Cancel crop tool -> Crop selection is removed when switching to any other tool. 
3. Switching to matching shape tool after selection of a shape with select tool -> Selection is kept
4. Switching to another shape tool after selection of a shape with select tool -> Selection is cleared
5. Switching to another shape tool after selection of a shape with matching tool  -> Selection is cleared
6. Selecting different shape types (resizable/not-resizable) -> resize nodes are properly cleared. 
7. _with switch to select tool option active_: Draw any shaped -> Selection is kept and select tool is activated.

### DIFF
--- a/ShareX.ScreenCaptureLib/Shapes/ShapeManager.cs
+++ b/ShareX.ScreenCaptureLib/Shapes/ShapeManager.cs
@@ -72,7 +72,6 @@ namespace ShareX.ScreenCaptureLib
             {
                 if (currentTool == value) return;
 
-                ShapeType previousTool = currentTool;
                 currentTool = value;
 
                 if (Form.IsAnnotationMode)
@@ -96,9 +95,12 @@ namespace ShareX.ScreenCaptureLib
                 if (CurrentShape != null)
                 {
                     // do not keep selection if select tool does not handle it
-                    if (currentTool == ShapeType.ToolSelect && !CurrentShape.IsHandledBySelectTool)
+                    if (currentTool == ShapeType.ToolSelect)
                     {
-                        DeselectCurrentShape();
+                        if (!CurrentShape.IsHandledBySelectTool)
+                        {
+                            DeselectCurrentShape();
+                        }
                     }
                     // do not keep selection if we switch away from a tool and the selected shape does not match the new type
                     else if (CurrentShape.ShapeType != currentTool)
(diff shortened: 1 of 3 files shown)


Danielku15:
@Jaex I really hope I covered this time all scenarios to make the select tool unselect/reselect as expected 😉
Simply let me know if you discover something more.

ShareX.ScreenCaptureLib/Shapes/ShapeManager.cs
```diff
@@ -93,9 +93,18 @@ public ShapeType CurrentTool
                     ClearTools();
                 }
 
-                if (previousTool != ShapeType.ToolSelect && currentTool != ShapeType.ToolSelect)
```
Jaex: You declared previousTool but it is not used anywhere?

Jaex:
Normally when you draw shape, that shape becomes selected. But if "Switch to selection tool after shape drawing" option is enabled then shape not stays selected. Is this intentional?

ShareX.ScreenCaptureLib/Shapes/ShapeManager.cs
Danielku15: True, it became obsolete with the rework on this delect logic. I will remove it.

Danielku15:
> Normally when you draw shape, that shape becomes selected. But if "Switch to selection tool after shape drawing" option is enabled then shape not stays selected. Is this intentional?

Auto-Switching to the select tool should keep the selection of the shape. It is only unselected if the drawn shape type is not handled by the select tool (e.g. the crop shape). [See here](https://github.com/Danielku15/ShareX/blob/e3b39685a00871616373aac501b7677fd9caf8d2/ShareX.ScreenCaptureLib/Shapes/ShapeManager.cs#L99)

But I fear the conditions are not 100% correct on this. The `!CurrentShape.IsHandledBySelectTool` part should go into the first if-branch as nested condition to not jump to the else-if part that unselects on generic tool switching.

Danielku15:
Sorry for the delay on the last fix. All scenarios of above are now covered.

Jaex:
Thanks.

PR #4173, opened 2019-06-15
Make settings accessible from the outside

### LINKED ISSUE #4172: Loading settings programmatically from the outside?
If I make my own Visual Studio solution with a Windows Forms project, reference all ShareX projects by adding the ShareX `csproj` files as existing projects and run the following code:

```
RegionCaptureTasks.GetRectangleRegion(out var rectangle, GetRegionCaptureOptions());
ScreenRecordManager.StartStopRecording(ScreenRecordOutput.FFmpeg, ScreenRecordStartMethod.CustomRegion, new TaskSettings()
{
    CaptureSettings = new TaskSettingsCapture()
    {
        FFmpegOptions = new FFmpegOptions()
        {
            VideoCodec = FFmpegVideoCodec.libvpx,
            VPx_bitrate = 2000
        },
        CaptureCustomRegion = rectangle
    }
});
```

Then it fails because `Program.Settings` is `null` (null reference exception). However, `SettingManager` and `Program.Settings` are `internal`, so I can't access them from my project.

Saving images works just fine, but I need video capture functionality as well in my application.

Can we make the `SettingManager` exposed to the outside (making it public), or perhaps make `Program.Settings` public instead?

I made a PR for my proposal of fixing the issue.

ffMathy:
Fixes #4172

### DIFF
--- a/ShareX/Program.cs
+++ b/ShareX/Program.cs
@@ -88,7 +88,8 @@ namespace ShareX
         public static bool IgnoreHotkeyWarning { get; private set; }
         public static bool PuushMode { get; private set; }
 
-        internal static ApplicationConfig Settings { get; set; }
+        public static ApplicationConfig Settings { get; set; }
+
         internal static TaskSettings DefaultTaskSettings { get; set; }
         internal static UploadersConfig UploadersConfig { get; set; }
         internal static HotkeysConfig HotkeysConfig { get; set; }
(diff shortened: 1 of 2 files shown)


ffMathy:
@Jaex let me know if the above doesn't make sense and I need to elaborate further. Once again thanks in advance.

PR #4200, opened 2019-06-30
Add option to keep image's aspect ratio

n3oney:
If width or height set to 0 using PercentageOfCanvas, the image will keep the aspect ratio to fit to the


StarToLeft:
Interesting, I'd love this to be implemented!

PR #4239, opened 2019-07-28
Add Persian Language Support

pourmand1376:
Hi, I've added Persian (Iran) to ShareX.


Jaex:
You translated like 2% part of whole application, please pull request after translation is 100%. Thanks.

PR #4258, opened 2019-08-09
Update BackblazeB2.cs

### LINKED ISSUE #4187: b2 service_unavailable pod is too busy
```
B2 uploader: ParseB2Error() got: 
{
  "code": "service_unavailable",
  "message": "c002_v0001118_t0002 is too busy",
  "status": 503
}
```
On occasion, the b2 uploader will fail due to service unavailable 503 error, and fail to request a new URL.

According to the linked reddit post, this is by design at backblaze. 

> When using Backblaze, a 503 error from time to time is EXPECTED AND NORMAL. We are thinking about doing a blog post about this, because we get your EXACT question in support from time to time and it seems to confuse customers.

> The "API software contract" with Backblaze says that if the pods you are uploading to tell you that they are "too busy" you are supposed to ask for a new upload URL (to a different pod) and upload to that different pod.

> The most common reason a pod might ask you to upload to a different pod is if the vault the pod is part of (a group of 20 pods is a vault) is now "full" and you need to upload to some OTHER vault. This isn't an error, it is literally expected behavior.

> All software that talks to B2 is supposed to handle this.

https://www.reddit.com/r/backblaze/comments/9273jx/b2_is_too_busy/e34g80x/

atajsic:
Fixes #4187

### DIFF
--- a/ShareX.UploadersLib/FileUploaders/BackblazeB2.cs
+++ b/ShareX.UploadersLib/FileUploaders/BackblazeB2.cs
@@ -213,6 +213,12 @@ namespace ShareX.UploadersLib.FileUploaders
                     DebugHelper.WriteLine("B2 uploader: Too Many Requests, trying with same URL.");
                     continue;
                 }
+                else if (uploadResult.RC == 503)
+                {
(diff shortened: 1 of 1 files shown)


Jaex:
Thanks.

PR #4277, opened 2019-08-23
A bit modification of about form

T3RY4:
Here I tried to edit about form more stylish. So whats do you think?

### DIFF
--- a/ShareX/Forms/AboutForm.cs
+++ b/ShareX/Forms/AboutForm.cs
@@ -27,7 +27,6 @@ using ShareX.HelpersLib;
 using ShareX.Properties;
 using System;
 using System.Windows.Forms;
-using System.Runtime.InteropServices;
 
 namespace ShareX
 {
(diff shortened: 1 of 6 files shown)


ShareX/Forms/AboutForm.cs
```diff
@@ -27,13 +27,32 @@
...
+using System.Runtime.InteropServices;
 
 namespace ShareX
 {
     public partial class AboutForm : Form
     {
         private EasterEggAboutAnimation easterEgg;
 
+        [DllImportAttribute("user32.dll")]
+        public static extern int SendMessage(IntPtr hWnd, int Msg, int wParam, int LPAR);
```
sylveon: Use the NativeMethods class.

T3RY4:
Screenshot of modification about form:[https://imgur.com/84irODE](url)

T3RY4:
I tested and now I cant find any bugs with this feature.

Jaex:
Sorry but I prefer to keep title bar of that form for consistency with other forms.

PR #4298, opened 2019-08-28
catching all Exceptions!

pourmand1376:
It's just a simple change to catch all exceptions in program because when this is missed, sometimes exceptions are not catched properly. 

[StackOverFlow Post about this](https://stackoverflow.com/questions/5762526/how-can-i-make-something-that-catches-all-unhandled-exceptions-in-a-winforms-a)

### DIFF
--- a/ShareX/Program.cs
+++ b/ShareX/Program.cs
@@ -237,6 +237,7 @@ namespace ShareX
 #if !DEBUG // Allow Visual Studio to break on exceptions in Debug builds.
             Application.ThreadException += Application_ThreadException;
             AppDomain.CurrentDomain.UnhandledException += CurrentDomain_UnhandledException;
+            Application.SetUnhandledExceptionMode(UnhandledExceptionMode.CatchException);
 #endif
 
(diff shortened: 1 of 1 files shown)


Jaex:
Which exception did not catch by ShareX?

pourmand1376:
Honestly, I don't know exactly why we should call this method, It's just recommended by Microsoft.

As Microsoft says : [(Unhandled Exception)](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.application.setunhandledexceptionmode?view=netframework-4.8) 

> It is often not feasible to catch all of the exceptions thrown by Windows Forms. Using this method, you can instruct your application whether it should catch all unhandled exceptions thrown by Windows Forms components and continue operating, or whether it should expose them to the user and halt execution.

> Call SetUnhandledExceptionMode before you instantiate the main form of your application using the Run method.

Jaex:
I mean did you encounter not catched exception in ShareX, is that why you decided to pull request this?

pourmand1376:
No, I did it because I've been doing this as best practice in all applications I participated in.

PR #4314, opened 2019-08-29
Adding Persian Language Support

pourmand1376:
Hi, I've added Persian Language support. Note that I've also added **fa-IR** designers in **csproj** file if you don't mind. 

Thank you for your consideration in advance.


Jaex:
Thank you for new translation!

Jaex:
> فارسی (Iran)

Isn't this should be "Persian" instead of "Iran" in enum list?

pourmand1376:
Yes, You're right. It should be Persian. 

Thanks for merging.

PR #4321, opened 2019-08-30
Update french translation

simonporte:
Hello,
I updated the french translation, the latest build brought new untranslated strings.
I noticed some strings that are not present in english but in french.
I framed them in the picture below.
Should I blank them ?

image

Thanks.


Jaex:
Thanks for translations.

When I remove controls from form, english entries deleting automatically but all other languages keep their entries. Manually finding all of them and removing would be too much work. So I just keep them.

PR #4387, opened 2019-09-23
update zh-TW localization

candela97:
Updated zh-TW localization files and created the ones that didn't exist, let me know if I missed anything!


Jaex:
Thanks.

Issue #820, opened 2015-07-23
Delete metadata before uploading support


vavavr00m:
Is this already possible with ShareX 10.0, if yes, how, if not, will it be implemented?

I did some searching, which, "hopefully", may be of use:
Image:
[ExifTool](http://www.sno.phy.queensu.ca/~phil/exiftool/)
[Metadata Extractor](https://github.com/drewnoakes/metadata-extractor-dotnet)
[Exiv2](http://www.exiv2.org/index.html)
Audio:
[TagLib](https://github.com/taglib/taglib)
[Quodlibet](https://github.com/quodlibet/quodlibet)
[Tags Library](http://www.3delite.hu/Object%20Pascal%20Developer%20Resources/TagsLibrary.html)

Jaex:
You can use actions to run CLI before upload.

vavavr00m:
I meant with the files from phone/DSLR or photo manipulators. I'm using ShareX not just for screenshots but also for regular uploading :-) It's really helpful.

Wferr:
Would love this to be implemented.

josh-richardson:
If this could be supported it would be great! Thanks!

Jaex:
https://x.com/ShareX/status/1920392978343833909

dumbu3:
Hello, how do i make it like it removes metadata after a screenshot / video?

PR #4471, opened 2019-11-03
#820: Added support for complete removal of image metadata on upload.

Arkanion:
This will remove all EXIF/XMP and other metadata from an Image.
A new option was added "RemoveUploadedImageMetadata" with default value false.

ImageHelpers.StripImageMetadata() will create a copy of the original files metadata information and blanks each value. 
Afterwards a new file is created with the stripped metadata, which will replace the original file.

This will not result in any image quality degredation or loss of information as the image is not reencoded, everything happens InPlace.

### DIFF
--- a/ShareX.HelpersLib/Helpers/ImageHelpers.cs
+++ b/ShareX.HelpersLib/Helpers/ImageHelpers.cs
@@ -34,6 +34,8 @@ using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Text;
 using System.Windows.Forms;
+using System.Windows.Media.Imaging;
+using ShareX.HelpersLib.Properties;
 
 namespace ShareX.HelpersLib
 {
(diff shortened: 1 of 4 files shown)


ShareX.HelpersLib/Helpers/ImageHelpers.cs
```diff
@@ -2022,5 +2024,123 @@ public static Size GetImageFileDimensions(string filePath)
 
             return Size.Empty;
         }
+
+        public static bool StripImageMetadata(string fileName)
+        {
+            bool result = false;
+
```
AraHaan: I would remove this space; well when I do my projects I do.

ShareX.HelpersLib/Helpers/ImageHelpers.cs
```diff
@@ -2022,5 +2024,123 @@ public static Size GetImageFileDimensions(string filePath)
...
+                        case ".tif":
+                            output = new TiffBitmapEncoder();
+                            break;
+                    }
+
+                    if (original.Frames[0]?.Metadata != null)
+                    {
+                        BitmapFrame copyFrame = (BitmapFrame)original.Frames[0].Clone();
+                        BitmapMetadata metaData = original.Frames[0].Metadata.Clone() as BitmapMetadata;
+
```
AraHaan: These spaces that are not after any ``}`` seems unnecessary for me. could remove these too.

Also helps with keeping things short and readable.

ShareX.HelpersLib/ShareX.HelpersLib.csproj
```diff
@@ -84,6 +84,7 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="PresentationCore" />
```
AraHaan: Why WPF binaries? Cant the metadata be cleared using non-WPF binaries too?

ShareX/WorkerTask.cs
```diff
@@ -398,6 +398,17 @@ private void DoUploadJob()
 
                 if (!cancelUpload)
                 {
+                    if (Program.Settings.RemoveUploadedImageMetadata)
+                    {
+                        string[] supportedExtensions = { @".jpg", @".jpeg", @".png", @".tiff" };
```
AraHaan: What about extensions that are provided via windows OS codecs that extends onto the WIC (Windows Imaging Component)? Should these be added to the supportedExtensions array too if found?

skyhirider:
This looks like an interesting feature, will it be updated and merged?

AraHaan:
Why is this so dead?

AraHaan:
F

PR #4491, opened 2019-11-18
Smoothened regionAnimation rectangle animation

Zaczero:
Whenever you move your mouse quickly around the shapes on the screen there are usually some visual glitches as the FromRectangle value is always set to PreviousHoverRectangle. This simple change smoothens those animations and improves the overall user experience. Simply use the last CurrentRectangle as the new FromRectangle value.

Preview before/after: https://i.imgur.com/WI0Gof7.mp4


Jaex:
Thanks.

Issue #2292, opened 2017-02-15
[Request] Copy multiple tagged urls/forum code urls from history panel


tioaxel:
When dragging and dropping for a batch upload, the files are usually out or order and cannot be sorted by filename:
[http://i.imgur.com/8Cruzzw.png](url)
This shouldn't be an issue because history lets me sort and filter, but here's the kicker: on Sharex's activity window, I can select multiple images, right click -> Copy -> Forum (BB Code) Linked Image (for example), and it'll copy the Forum (BB Code) Linked Image for all selected images. In the History window, this doesn't work (it only copies the code for the first selected file); the only 'multiple copy' options that work are URL and HTML Link.

So, my request: make copying multiple link/image/linked images available for history, or make sorting available for the activity window.

Thank you.

Jaex:
Added by #4548

PR #4548, opened 2020-01-01
Added the ability to mass copy links in different formats

C03:
This fulfills the request to implement the ability to mass copy links when multiple items are selected in the 'history' area https://github.com/ShareX/ShareX/issues/2292.

### DIFF
--- a/ShareX.HistoryLib/HistoryItemManager.cs
+++ b/ShareX.HistoryLib/HistoryItemManager.cs
@@ -161,10 +161,9 @@ namespace ShareX.HistoryLib
 
         public void CopyURL()
         {
-            if (HistoryItem != null && IsURLExist)
+            HistoryItem[] historyItems = OnGetHistoryItems();
+            if (historyItems != null)
             {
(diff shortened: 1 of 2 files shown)


Jaex:
It is bad idea to use `OnGetHistoryItems().Length` in that function, especially multiple times. Also you not let opening URL when there is multiple items selected but currently it at least supports opening first URL.

C03:
Hello, thanks for your response. 

I'm open for suggestions on an alternative option to `OnGetHistoryItems().Length` but personally I'm unsure on why it's bad idea to use it. You are completely right about multiple uses, I should have used a local variable to get the count and then used that instead.

As for the opening URL options, I can add it so that it opens all URLs if that would be appropriate? My concern was that if someone had selected 50 URLs and pressed the open option it could cause their browser to crash.

Jaex:
Thanks.

Jaex:
Did you had any reason to not support multi copying in these last 4 buttons?

C03:
I've now added that in https://github.com/ShareX/ShareX/pull/4555

PR #4555, opened 2020-01-03
Added support for mass copying paths

C03:
Added support for mass copying:
-Folder paths
-File paths
-File names
-File names with extensions


Jaex:
Thanks!

Issue #4521, opened 2019-12-11
Hide all "Upload File" Commands in UI


megabugman:
I don't plan to ever upload anything, therefore I set DisableUpload "True" in Settings.
In Image history window the command "upload file" is still visible. Also the main Window 
still shows the Upload Button containing the subcommands Upload File/Folder,etc.

I miss an option to hide everything (buttons, menu entries, action toolbar, etc) regarding uploading in the UI including OCR, basically a offline version. ATM I see a big chance of unmeant data leakage. Thanks

Jaex:
> In Image history window the command "upload file" is still visible. Also the main Window
still shows the Upload Button containing the subcommands Upload File/Folder,etc.

Both history and main window upload buttons does not upload if you enabled "DisableUpload" option.

> including OCR

Now I made sure "DisableUpload" option also gonna disable OCR functionality in next release.

> ATM I see a big chance of unmeant data leakage.

If you enabled "DisableUpload" option then only OCR bypass that option. So there is no big chance of data leakage. If there is any other place which not respects that option then please let me know.

> basically a offline version

Technically you can accomplish this for most applications by blocking them using firewall software.

PR #4556, opened 2020-01-05
Added an "offline mode"

C03:
Implemented an "offline mode" for https://github.com/ShareX/ShareX/issues/4521
-Adds "OfflineMode" in Application Settings > Advanced > Application 
-Hides everything requiring connectivity (OCR, Uploading, Debug, Proxy etc)

The only issue I had with this was 3 items on the "After Capture Tasks" drop-down menu:

If you could tell me where these menu items are located then I'll be able to hide them as well.

### DIFF
--- a/ShareX.HelpersLib/HelpersOptions.cs
+++ b/ShareX.HelpersLib/HelpersOptions.cs
@@ -40,5 +40,6 @@ namespace ShareX.HelpersLib
         public static List<Color> RecentColors { get; set; } = new List<Color>();
         public static string LastSaveDirectory { get; set; } = "";
         public static bool URLEncodeIgnoreEmoji { get; set; } = false;
+        public static bool OfflineMode { get; set; } = false;
     }
 }
--- a/ShareX.HistoryLib/HistoryItemManager_ContextMenu.cs
+++ b/ShareX.HistoryLib/HistoryItemManager_ContextMenu.cs
@@ -23,6 +23,7 @@
 
 #endregion License Information (GPL v3)
 
+using ShareX.HelpersLib;
 using ShareX.HistoryLib.Properties;
 using System;
 using System.Drawing;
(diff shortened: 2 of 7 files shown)


C03:
Apologies for the styling issues

Jaex:
I don't see point of this option. If you don't want to upload just don't press those buttons. And if it is bothering anyone, people are feel free to use screen capture software which have no upload capabilities, such as built in "Screen snip" of Windows 10.

You may tell what if users accidentally press upload buttons, thats why I added "DisableUpload" option, which will make sure upload not executes.

C03:
"DisableUpload" doesn't disable OCR? The user was requesting that the application would run in a way that their images couldn't accidentally be uploaded. 

Sure this feature may be not widely used but that's why it's hidden away in the advanced settings - the purpose of this feature is for those who work in sensitive environments where accidental data leakage can have some pretty bad effects depending who manages to get access to that data. 

The suggestion of using "Snip & Sketch" is invalid for the following reasons:
-It's for Windows 10 only
-No GIF support
-Editor is mediocre
-Lacks the ability to auto copy the image to the clipboard

But if you still disagree with this feature then could you at least close the issue requesting this feature?

sylveon:
Unfortunately, ShareX does not support corporate environments really well.

The lack of group policy controls means for example any user can disable `DisableUpload` and upload private information away, and as you've noticed things like OCR ignore that entirely despite uploading your data to a service.

The issue I see here, and probably what @Jaex means, is that your offline mode is almost a duplicate for `DisableUpload`. Why not make your PR improve that to also hide features that upload, instead of adding an entirely new setting?

C03:
Thanks for the response, that's a good shout. I can implement this into `DisableUpload` option rather than having it as a separate feature. 

In regards to corporate environments, I was more aiming this at users who work as freelancers where accidentally uploading pictures of their code/designs can actually be a breach their contract.

@Jaex does this sound like a reasonable change?

PR #4622, opened 2020-02-19
Add alternative default ffmpeg location under install folder

### LINKED ISSUE #4621: ffmpeg global install location
First of all, thanks for the great tool!

I didn't find the option to install ffmpeg globally for all users on computers. I thought that appropriate location would be "[install dir]\Tools\ffmpeg.exe" but it doesn't work. An use case would be when you want to mass deploy ShareX and you also want out-of-the box screen recording capability - without the need that user has to download ffmpeg.exe from Internet.

Will send PR for this ...

schtritoff:
This PR fixes #4621, adds posibility to install ffmpeg globally in install dir. Use case is for example to first install sharex silently in batch script and then copy ffmpeg.exe to %programfile%\ShareX\Tools\ffmpeg.exe

### DIFF
--- a/ShareX/Program.cs
+++ b/ShareX/Program.cs
@@ -233,7 +233,19 @@ namespace ShareX
 
         public static string ToolsFolder => Path.Combine(PersonalFolder, "Tools");
         public static string ScreenRecorderCacheFilePath => Path.Combine(PersonalFolder, "ScreenRecorder.avi");
-        public static string DefaultFFmpegFilePath => Path.Combine(ToolsFolder, "ffmpeg.exe");
+        public static string DefaultFFmpegFilePath
+        {
+            get
+            {
+                string ffmpegFilePath = Helpers.GetAbsolutePath("Tools\\ffmpeg.exe");
+                if (!File.Exists(ffmpegFilePath))
+                {
+                    ffmpegFilePath = null;
+                }
+
+                return ffmpegFilePath ?? Path.Combine(ToolsFolder, "ffmpeg.exe");
+            }
+        }
         public static string ChromeHostManifestFilePath => Path.Combine(ToolsFolder, "Chrome-host-manifest.json");
         public static string FirefoxHostManifestFilePath => Path.Combine(ToolsFolder, "Firefox-host-manifest.json");
 


sylveon:
Is deploying a config file which points to a global ffmpeg install not possible?

ShareX/Program.cs
```diff
@@ -233,7 +233,19 @@ public static string ScreenshotsFolder
 
         public static string ToolsFolder => Path.Combine(PersonalFolder, "Tools");
         public static string ScreenRecorderCacheFilePath => Path.Combine(PersonalFolder, "ScreenRecorder.avi");
-        public static string DefaultFFmpegFilePath => Path.Combine(ToolsFolder, "ffmpeg.exe");
+        public static string DefaultFFmpegFilePath
+        {
+            get
+            {
+                string ffmpegFilePath = Helpers.GetAbsolutePath("Tools\\ffmpeg.exe");
+                if (!File.Exists(ffmpegFilePath))
```
sylveon: This seems uselessly complex.

```cs
if (File.Exists(ffmpegFilePath))
{
    return ffmpegFilePath;
}
else
{
    return Path.Combine(ToolsFolder, "ffmpeg.exe");
}
```

schtritoff: Well I think this is purely opinion based - I was following the rule that method/function should have only 1 return statement (https://stackoverflow.com/a/866092/1155121).

schtritoff:
> 
> 
> Is deploying a config file which points to a global ffmpeg install not possible?

Current implementation has hard-coded default value for per-user installation of ffmpeg so I only added option to have default value for per-machine ffmpeg installation. Since this method is called only on first start I don't think that global config would make much more difference.

Maybe It would make more sense to distribute (if license permits) ffmpeg with setup program or to make some other changes to setup - for example to copy/install ffmpeg to default global location if present during installation, etc.

That is all out of scope of this PR since I'm only trying to enable per-machine ffmpeg installation.

schtritoff:
@sylveon Since it looks like this is sliding into stale PR, do you have any suggestion how to approach this? The goal should be to enable ffmpeg installation at setup time (incorporating into inno), and not at application runtime as it is now. Also if possible not to require internet connection to setup ffmpeg.

sylveon:
Is using the Microsoft Store an option? This build has ffmpeg built-in due to store restrictions

schtritoff:
I'm trying offline install which excludes installing directly from Microsoft Store. But I tried with appx package available on AppVeyor, this command:
`Dism /Online /Add-ProvisionedAppxPackage /PackagePath:D:\tmp\DL\ShareX.appx /SkipLicense`
But got this error: 

> Error                 DISM   DISM Appx Provider: PID=1144 TID=13512 Failed to get signature footprint file for package 19568ShareX.ShareX_13.1.0.0_x64__egrzcvs15399j. - PackageMonikerToPayload::CreateForPackageReader(hr:0x80070002)

It there download url for official released appx version ShareX that is properly signed?

sylveon:
You can use this to get the Microsoft-signed APPX: https://store.rg-adguard.net/

Put the ShareX store link (https://www.microsoft.com/store/apps/9nblggh4z1sp), select the retail branch, and press the checkmark.

schtritoff:
After downloading appx from Microsoft Store this is actually working as expected. Only downside is that there is no x86 build but that's fine for my case. 

Thanks

PR #4688, opened 2020-03-19
Update french translation

simonporte:
Complete missing strings
Correct end of string ponctuation
Plus a few adjustments for word consistency


Jaex:
Thanks.

PR #4758, opened 2020-04-21
Replace System.Remoting by named pipe, remove RIPEMD-160 support

sylveon:
This is a step towards full .NET Core API coverage. It makes ShareX.HelpersLib 100% covered by .NET Core + Platform Extensions, from a 99.41% previously

Before: 
image
image

After:
image
image


sylveon:
Now 100% covered

image

PR #4774, opened 2020-04-29
PT-PT Language

FarewellAngelina:
Translated all strings to PT-PT Language, please pull this.
Thank you.


Jaex:
Thanks for translation! But these should not be included:

Jaex:
I noticed pt-pt files missing in some places like here:

FarewellAngelina:
What path is that from? Because i see them on ShareX/ShareX.HelpersLib/Properties/ , do you have more examples?

Jaex:
Interesting for some reason that specific single file did not show up in Visual Studio when all other resx files showed up but now I can see it. Not sure why.

`Português (Portuguese)` for language menu right?

FarewellAngelina:
Yes, that's right.

Issue #4746, opened 2020-04-13
Turn off right click to exit Sharex Image Editor


Ramadas108:
Now right click in Sharex Image Editor makes it shut down without prompt for saving. It is very annoying because it can be done by mistake.... and work is lost

livan3li:
yes i was looking for a solution for this issue. How can we overcome it. İ am drawing, typing cropping etc and accidently a right click, all of them is gone! 

The exit choice can be Escape(ESC key) in my opinion no necessery ESC and Right click. One is sufficiant

Obi-Sean:
Just had this betray me yesterday, and even updating to v17 this morning has not solved it. Can we *please* at least get a work around on this?

PR #4781, opened 2020-04-30
fixed issue #4746: Right clicking while on region capture will show a confirmation msg before closing

lloyd99901:
ShareX will now prompt the user if they want to close down the region capture/image editor after pressing right-click.


Jaex:
It already does:

Jaex:
I noticed now both esc key press or trying to close window with X button shows that dialog but not right click. So I will use same dialog we use already for right click too but only for image editor. Because region capture need to be closed quickly with right click.

Jaex:
This is the bug: https://github.com/ShareX/ShareX/commit/b5ccbd612eccee5c489acdd46b015d97e7a19de5

PR #4784, opened 2020-05-01
Added a right-click context menu on the screenshot preview (NotificationForm).

lloyd99901:
Added a right-click context menu which contains delete button, open file/file location, copy file/file location, etc, on the screenshot preview (NotificationForm).
ShareX_k4561Qvoja
ApplicationFrameHost_8AkgIoPCwc

### DIFF
--- a/ShareX/Enums.cs
+++ b/ShareX/Enums.cs
@@ -2,6 +2,31 @@
 
 /*
     ShareX - A program that allows you to take screenshots and share any file type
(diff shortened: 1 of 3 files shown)


JLiszt:
I know I'm late, but this seems like a great addition and would be cool if it were added.

lucario:
If you can get into contact with @lloyd99901 , they will have to make a new pr, or someone else.

PR #4786, opened 2020-05-02
Added check box to make direct URL for SeaFile uploader

user6323:
Now ShareX create link for SeaFile uploader which lead to SeaFile page.
The problem that if the image is larger than the available space on the SeaFile page than the image is shrunk. There is no ability to view it in full size without downloading first, what is inconvenient.

But there is the ability to create a direct link by adding "raw=1" parameter for the url.
Example:
"http://example.com/f/af416996a8064ada9c7c/" - leads to SeaFile page.

"http://example.com/f/af416996a8064ada9c7c/?raw=1" - leads to file directly.

I added a checkbox called "Direct URL" to the SeaFile uploader configuration so that the parameter "raw" is added to the address. 
ShareX_YQHzMiVxWB

Now it is possible to get direct link right after uploading.

### DIFF
--- a/ShareX.UploadersLib/FileUploaders/Seafile.cs
+++ b/ShareX.UploadersLib/FileUploaders/Seafile.cs
@@ -58,7 +58,7 @@ namespace ShareX.UploadersLib.FileUploaders
                 ShareDaysToExpire = config.SeafileShareDaysToExpire,
                 SharePassword = config.SeafileSharePassword,
                 CreateShareableURL = config.SeafileCreateShareableURL,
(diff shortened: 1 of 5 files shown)


Jaex:
Checkbox text of "Create shareable URL" can overlap your checkbox in some translations.

Calling it "Use raw URL" can be more valid I think? Because it redirects to direct url and not actually direct url.

user6323:
Renamed variables and titles, and all elements below the new checkbox moved down to fit the new checkbox.
ShareX_xLAsTId7GN

Jaex:
Thanks.

PR #4796, opened 2020-05-07
Refresh action icon only on context menu call and only once per path change

user6323:
There is a visible delay on every click and right click on upload element before the preview is shown.

I have 6 external programs added as actions. I figured out what on every click program loading every image for every external program.
The more these programs, the longer is delay.

At first, i am storing received bitmaps in ExternalProgram model. ShareX now can reuse it. If the path changes then ShareX will retrieve new images with changed path.

Second, looks like the icons for external images used only in context menu, in "Run action" submenu.
So, action menu now filled only on right mouse click, or Keys.Apps key (used for keyboard context key).

Third, i moved visibility control of tsmiRunAction inside a single method because i encounter a bug about wrong visibility when i open context menu one by one very fast. This change fixes this.

Now thumbnails loaded almost instantly on click:

### DIFF
--- a/ShareX.HelpersLib/ExternalProgram.cs
+++ b/ShareX.HelpersLib/ExternalProgram.cs
@@ -25,6 +25,7 @@
 
 using System;
 using System.Diagnostics;
+using System.Drawing;
 using System.IO;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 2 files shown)


Jaex:
Actions menu not showing up in multiple places for example in thumbnail view:



Also storing Icon in that class is not good idea because it is saved as setting or cloned for workflows etc. so even it is not serialized still because Icon is not needed for workflows and only used for main window, therefore it is still not good idea I think.

Instead of separating actions menu logic from UpdateInfoManager(), need to separate whole context menu logic from that function I think to decrease delays after selection changes.

I saw you used two different casing for private variables here:



One starts with _ and other starts with upper case. Where did you saw in this project that private variables starts with _ character? Don't you think project should have consistent naming?

Jaex:
And now while opening action menu I received this exception:

Jaex:
Also now I noticed that "Run action" button shows up even when it have no sub menu:

PR #4810, opened 2020-05-15
Allow "access level" selection for Box shareable link

fib25:
Adds a dropdown box to the Box file uploader settings under the "Create shareable link" checkbox.

sharex_box_share_link_access_level

This has been tested with all 3 access levels and I saw each successfully upload with the correct share link setting in the Box web interface.

See Box API for reference about access levels: https://developer.box.com/guides/shared-links/create/
**Please note** the "collaboration" access level in the Box docs should actually be "collaborators" as seen in their own API implementations (e.g. https://github.com/box/box-node-sdk/blob/b06f4041d54fb2b102603edd2624876d25500e9a/lib/box-client.js#L392).
I have tested and made sure I to add the correct "collaborators" value.

### DIFF
--- a/ShareX.UploadersLib/Enums.cs
+++ b/ShareX.UploadersLib/Enums.cs
@@ -353,4 +353,14 @@ namespace ShareX.UploadersLib
         Unlisted,
         Private
     }
+
+    public enum BoxShareAccessLevel
+    {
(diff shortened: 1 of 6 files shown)


Jaex:
Thanks.

PR #4886, opened 2020-06-17
Set anchor of label to the bottom-left in TextDrawingInputBox

morsisko:
Before the commit:

After the commit:


Jaex:
You needed to just change anchor of control, not dock it.

morsisko:
Guess I can't change the content of this PR? Anyway as far as I see the label acts the same (at least visually), no matter if docked or with changed anchor of control.

lucario:
@morsisko Just push the change to the same branch.

morsisko:
Ok, I've changed the dock to "Bottom, Left" anchor

Jaex:
Why did you move control to left.

morsisko:
Sorry, I've ammended the last commit so it is now only plain anchor change to the bottom-left (because as I understand this is the intended position for the label)

PR #4935, opened 2020-07-18
Allow quick tasks to use a specific image effect preset

ledshok:
As a ShareX user my workflow revolves around the use of a single hotkey (PrintScr) and a default image effect (add a shadow). However, there are occasions where I'd like a different combination of effects to be applied (eg shadow + resize, or resize but no shadow, etc). 

While this could be achieved using multiple hotkeys, I find having a single hotkey a lot simpler to remember, and thus quick tasks come to the rescue!

This change allows (if the application settings is enabled) the user to select from a list of default task image effect presets when configuring a quick task. 

Here are my three default task image effect presets:
image

And here they appear in a sub-menu (under 'Add image effects') when configuring a quick task:
image

When the hotkey is pressed and the hotkey workflow pops the quick task menu, the user can effectively choose which of the image effect presets should be applied:
image

It seems at least one other ShareX user is looking for similar functionality - https://www.reddit.com/r/sharex/comments/ghujxg/can_you_assign_different_image_effects_to/

Note: I'm just a hobbyist developer, so apologies in advance if my changes are a bit amateurish!

### DIFF
--- a/ShareX/ApplicationConfig.cs
+++ b/ShareX/ApplicationConfig.cs
@@ -186,6 +186,9 @@ namespace ShareX
         [Category("Application"), DefaultValue(false), Description("In main window when task is completed automatically select it.")]
         public bool AutoSelectLastCompletedTask { get; set; }
 
+        [Category("Application"), DefaultValue(false), Description("Allow Quick Tasks to use specific image effect presets.")]
+        public bool AllowQuickTaskImageEffectPresets { get; set; }
+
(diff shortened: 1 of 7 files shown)


Lkuijk:
This would be very awesome. I am a writer of knowledge bank articles and I very a lot in image effects on the screenshots. Should save a lot of time if I can have a shortcut for every effect.

Unknow0059:
This should be in sharex already.

PR #4954, opened 2020-07-31
Added Copy, Save and SaveAs hotkeys in Capture mode

ya-erm:
These changes allows you to use hotkeys while capturing region.
I made it to improve UX. 
User case before changes: Press PrntScrn → Select regions → Press Enter → Press Ctrl+Shift+S to save
After these changes: Press PrntScrn → Select regions → Press Ctrl+S to save
As you can see users can skip one step now and choose action (save or copy) by pressing hotkey.

Actions
Copy: Ctrl+C
Save: Ctrl+S
SaveAs: Ctrl+Shift+S

### DIFF
--- a/ShareX/CaptureHelpers/CaptureRegion.cs
+++ b/ShareX/CaptureHelpers/CaptureRegion.cs
@@ -26,6 +26,7 @@
 using ShareX.HelpersLib;
 using ShareX.ScreenCaptureLib;
 using System.Drawing;
+using System.IO;
 using System.Windows.Forms;
 
 namespace ShareX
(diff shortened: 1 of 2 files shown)


ShareX/CaptureHelpers/CaptureRegion.cs
```diff
@@ -87,6 +88,47 @@ protected ImageInfo ExecuteRegionCapture(TaskSettings taskSettings)
...
+                {
+                    using (output)
+                    {
+                        if (string.IsNullOrEmpty(newFilePath))
+                        {
+                            string fileName = TaskHelpers.GetFilename(taskSettings, taskSettings.ImageSettings.ImageFormat.GetDescription(), output);
+                            newFilePath = Path.Combine(taskSettings.GetScreenshotsFolder(), fileName);
+                        }
+
+                        newFilePath = ImageHelpers.SaveImageFileDialog(output, newFilePath);
```
HumanRV, line 117: the difference in one line. could it be parametrized?

ya-erm: it's just copied from method AnnotateImage in file TaskHelpers.cs https://github.com/ShareX/ShareX/blob/a883f2bf63aaf5a30182fc61299a8d9da3087064/ShareX/TaskHelpers.cs#L960-L998

PR #4966, opened 2020-08-05
Lossless optimization of images

tiagofilipesilva:
In this pull request several images used in ShareX's resources are optimized in a lossless way --- binary size shrinks, no quality loss.


tiagofilipesilva:
@Jaex This is a safe replace.

PR #4969, opened 2020-08-09
Added display path options to folder listing module.

JamieSharpe:
Added an additional feaature to the Directory Indexer module. There are now two settings to show the full path of the directories, or the shorter variation of the path starting from the root. Another update is to show number seperators on the folder and file count for clarity.

Comparison of ShareX directory indexing module changes

Notes in the picture:
Marker 1 - original root folder path.
Marker 2 - original sub folder path.
Marker 3 - original file/folder count.
Marker 4 - updated full folder path for root and sub folders.
Marker 5 - updated file/folder count with seperators.
Marker 6 - updated folder path limiting to the selected root folder.

### DIFF
--- a/ShareX.IndexerLib/IndexerHtml.cs
+++ b/ShareX.IndexerLib/IndexerHtml.cs
@@ -54,7 +54,7 @@ namespace ShareX.IndexerLib
             sbHtmlIndex.AppendLine(HtmlHelper.StartTag("body"));
 
             folderPath = Path.GetFullPath(folderPath).TrimEnd('\\');
-            prePathTrim = folderPath.LastIndexOf(@"\");
+            prePathTrim = folderPath.LastIndexOf(@"\") + 1;
 
             FolderInfo folderInfo = GetFolderInfo(folderPath);
             folderInfo.Update();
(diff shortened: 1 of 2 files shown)


Jaex:
If folder path ends with `\` then DisplayPathLimited option not gonna work correctly.

Also maybe not showing `\` at beginning of path can look better but not sure what users would prefer.

JamieSharpe:
> If folder path ends with `\` then DisplayPathLimited option not gonna work correctly.

Testing this out on the current stable release of ShareX (13.1), it seems this bug was present before the pull request. See Screenshot:

ShareX Directory Indexer Folder Path Seperator Bug


I've added a small modification to fix this potential issue.

> Also maybe not showing \ at beginning of path can look better but not sure what users would prefer.

I can add an option to show/hide the parent path seperator if need be.

Jaex:
> I can add an option to show/hide the parent path seperator if need be.

Do you need it? Because just not showing it looks better I think.

JamieSharpe:
> Do you need it? Because just not showing it looks better I think.

After comparing the two, it makes more sense to not have the path seperator at the start. It does look cleaner, and I can see a claim of confusion of it idicating a root folder of a partition; similar to Unix OS's.

The previous commit removes this prepended path seperator.

Jaex:
Thanks.

PR #4970, opened 2020-08-09
Fixed ruler disappearing when the region box is too small.

JamieSharpe:
When the ruler tool is opened, the region info window is hidden when the region is too thin, making the ruler useless for straight lines or accurate measurements.

This is a one line fix so the ruler is always displayed.

Ruler fix

Marker 1 - Test area
Marker 2 - Start position
Marker 3 - End position - No Information shown.
Marker 4 & 5 - Increase region size to display limit now shows the ruler information.
Marker 6: - Pull request fix to show the info window at any region size.


Jaex:
Click hover select not works after this PR:



So minimum size was to make sure if your hand shake while clicking still it can see it as click and not region drag.

PR #5017, opened 2020-09-03
Bayer Scale can be changed for gifs if Bayer dither is in use

Scrxtchy:
Adds a combobox allowing users to alter bayer_scale that is used when encoding a gif file with the bayer dithering mode.
The UI for it isn't that great, as in the label doesn't exist, however the Screen Recording menu is pretty packed as is, I have only put an indication of it's function within the tooltip regrettably.

image

There's a little bit of the bayer_scale in action on [this article](http://blog.pkh.me/p/21-high-quality-gif-with-ffmpeg.html#usage), otherwise it is covered in the [ffmpeg filters documentation](https://ffmpeg.org/ffmpeg-filters.html#paletteuse)


Jaex:
Thanks.

PR #5042, opened 2020-09-14
Added "Color depth" filter

somethingSTRANGE:
Added "Color depth" filter that remaps RGB pixel color to the nearest color at the specified color depth. For example, #E1CB96 will be remapped to #DDCC99 when "BitsPerChannel" is set to 4.

The attached image includes three sample images, each processed by the filter using the range of allowed bpc values (1-8).

 ColorDepth_Sample

### DIFF
--- a/ShareX.HelpersLib/Cryptographic/HashCheck.cs
+++ b/ShareX.HelpersLib/Cryptographic/HashCheck.cs
@@ -62,7 +62,7 @@ namespace ShareX.HelpersLib
                     {
                         return HashCheckThread(filePath, hashType, progress, cts.Token);
                     }
-                    catch (OperationCanceledException e)
+                    catch (OperationCanceledException)
                     {
                     }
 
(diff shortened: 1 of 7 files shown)


Jaex:
> catch (OperationCanceledException)

> return Task.Run(() => StartEncoding());

> <LangVersion>7.3</LangVersion>

Why did you change these?

somethingSTRANGE:
> > catch (OperationCanceledException)
> 
> > return Task.Run(() => StartEncoding());
> 
> > <LangVersion>7.3</LangVersion>
> 
> Why did you change these?

To fix the compiler issue I reported in the commit. The catch exception variable was unused (warning), and the Task.Run wouldn't compile due to an ambiguous call. The Enum generic constraints also would not compile without setting the Language version to 7.3.

It's possible these are caused by my compiler settings, but I'm not doing anything special. I first opened the project with Rider 2019 and received those issues when the solution failed to build. I then loaded the project up in a clean install of VS2017, and the same issues were reported there.

I even reverted those changes after making a build, and the compiler errors returned for me. The exception change was just to eliminate the unused variable warning.

Jaex:
>I'm not doing anything special

You are using very old version of Visual Studio probably which causing these problems.

PR #5071, opened 2020-09-24
Fix emoji encoding in URLs for FTP uploads

x04:
Currently if you use FTP/FTPS/SFTP to upload images it ignores the configuration option to ignore emoji's in the file name/URL. That makes it so I can't do things like [https://noyou.gay/🧣🧞👔🧦👵👈💁🚶.png](https://noyou.gay/🧣🧞👔🧦👵👈💁🚶.png), it just comes out to be a terrible URL encoded version of the emojis. With the changes in this PR that is fixed if they choose to ignore emojis.


Jaex:
I tested it now, it returned this URL:
`https://example.com/test test/test 😀 (🐱).txt`
So it not encodes non emoji parts. This is the reason why I did not add emoji support to FTP too.

x04:
That should fix the issues.

ShareX.HelpersLib/Helpers/URLHelpers.cs
```diff
@@ -35,6 +35,7 @@
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using System.Web;
+using System.Windows.Forms;
```
JakeSteam, line 38: 🤔

x04: Oh yeah I believe IDE autocompleted that when I was typing something, I'll remove it.

x04:
@Jaex any more input on this?

ha6000:
Bump

Issue #3934, opened 2019-02-04
Customizable color picker format


DrummerB:
It would be useful to be able to define custom export formats for the Color Picker tool, similar to the Screen Color Picker tool.

E.g. you could define export format that could be pasted directly into source code:

Format: `new Color($.3rf, $.3gf, $.3bf)` 
Example output: `new Color(0.123f, 0.456f, 0.789f)`

Format: `UIColor(red: $.2r, green: $.2g, blue: $.2b, alpha: $.2a)` 
Example output: `UIColor(red: 0.10, green: 0.25, blue: 0.50, alpha: 1.00)`

![image](https://user-images.githubusercontent.com/2104474/52209997-ec326d80-2885-11e9-8294-a8135881b78f.png)

FirstDagger:
This would be very useful _even if_ only an additional option for copy the hex value into the clipboard was available. Currently this is a deficiency of ShareXs IMHO.

Jaex:
Implemented in #5072

PR #5072, opened 2020-09-25
Update color picker format options to include 0-1 colors

JakeCover:
image
Basically what the title says, you can now use capital letters `$R`, `$G`, and `$B` to get a color string with colors represented from 0-1. This format of colors is used with a number of things, most notably Unity, and this will save me and hopefully others from having to convert normal colors to 0-1 range every so often, which I say is a win.    
I have it set to round to 3 decimal places, but that's totally arbitrary and any precision will basically work, although some level of rounding seems sensible otherwise you get unnecessarily long outputs. If you think it should be more/less precise, just LMK and I'll do whatever. Also, I think the use of `1.000` in the description strings gets the point across pretty well, but that can also be stated more clearly if need be. 
 
I looked around for a bit and couldn't find a contributor guide so I tried to just modify as little as possible and keep styling consistent. Not really sure what I'm supposed to be PRing into, I'm assuming master since most other recent commits are straight to master? I've tested this myself and it works fine, both capital and lowercase format characters work as intended, but feel free to verify yourself. Thanks for all the hard work on the best screenshot tool for windows! 
<sub><sub><sub>Also bankers rounding is a dumb default thanks Microsoft</sub></sub></sub>

### DIFF
--- a/ShareX.HelpersLib/NameParser/CodeMenuEntryPixelInfo.cs
+++ b/ShareX.HelpersLib/NameParser/CodeMenuEntryPixelInfo.cs
@@ -35,6 +35,9 @@ namespace ShareX.HelpersLib
         public static readonly CodeMenuEntryPixelInfo r = new CodeMenuEntryPixelInfo("r", "Red color (0-255)");
         public static readonly CodeMenuEntryPixelInfo g = new CodeMenuEntryPixelInfo("g", "Green color (0-255)");
         public static readonly CodeMenuEntryPixelInfo b = new CodeMenuEntryPixelInfo("b", "Blue color (0-255)");
+        public static readonly CodeMenuEntryPixelInfo R = new CodeMenuEntryPixelInfo("R", "Red color (0-1.000)");
+        public static readonly CodeMenuEntryPixelInfo G = new CodeMenuEntryPixelInfo("G", "Green color (0-1.000)");
+        public static readonly CodeMenuEntryPixelInfo B = new CodeMenuEntryPixelInfo("B", "Blue color (0-1.000)");
         public static readonly CodeMenuEntryPixelInfo hex = new CodeMenuEntryPixelInfo("hex", "Hex color value (Lowercase)");
         public static readonly CodeMenuEntryPixelInfo HEX = new CodeMenuEntryPixelInfo("HEX", "Hex color value (Uppercase)");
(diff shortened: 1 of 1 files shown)


Jaex:
Like this adding option for decimal places can be good idea:



By default it can be 3 decimal places.

About capital R G B, I used it for uppercase hex result so it can be inconsistent naming. Maybe using `$r255` for 0-255 and `$r1` for 0-1 can make more sense? `$r` can be still supported for backward compatibility.

Jaex:
Related issue: #3934

JakeCover:
Ok, said changes have been made! 
* It now supports variable decimal precision for `$r1{n}` substitutions, and will default to 3 if no `{n}` is provided. One issue is that the `Math.Round()` function only accepts numbers between 0 and 15 inclusive for place values to round to, so any value greater or smaller will be clamped to the range 0-15. 
* 0-255 range RGB will now be provided on `$r255`, as well as `$r` for backwards compatibility, however `$r` does not show up in the suggestion list to avoid unnecessary clutter (not sure if this is best, let me know what you think, but it seemed reasonable to me). Attached below is a screenshot to show what I mean: 

image

* I'm not totally sure what the best ordering of the list is, although this seems good enough to me. Just LMK if you think something else would be better. 

A good sample string to demonstrate most concepts is `$r1{-5}, $g1, $b1{13} | $r255, $g, $b{5}` ---> `0, 0.118, 0.1176470588235 | 30, 30, 30{5}`
* `$r1{-5}` - `Math.Round()` will not accept negitave numbers, so `n` of -5 becomes 0 and therefore this is rounded to a whole number. If `n` was larger than 15, it would be clamped down to 15 as well.
*  `$g1` - Default case, rounded to 3 decimal places.
* `$b1{13}` - Example case of `{n}`, just rounds to that many places.
* `$r255` - Works like old `$r` would, just returns the 0-255 repr.
* `$g` - Literally is the old `$g`, exists for backwards compatibility.
* `$b{5}` - Shows that brackets don't work on anything other than `$r1` case.

As a sidenote, I'm more of a python developer than a C# developer, so I might be doing some thing in less than optimal ways. If anything stands out, just let me know what a better approach would be and I'd be happy to change it.

Jaex:
Thanks, good job.

PR #5106, opened 2020-10-06
Allow drawing arrows in reverse

### LINKED ISSUE #5103: Modifier to draw arrows in reverse
Provide a modifier click and drag to draw an annotation arror in the opposite direction.

I find it more efficient (ie. requiring fewer subsequent touchups/corrections) to start drawing an arrow from the object to a coarsely placed area (which is where the legend will go) rather than end at the object (current fashion). Would be great if - for example - alt could be used to temporarily reverse the direction.

Scrxtchy:
Adds a shape option for drawing arrows pointing to their origin
Closes #5103

image
image

### DIFF
--- a/ShareX.HelpersLib/Enums.cs
+++ b/ShareX.HelpersLib/Enums.cs
@@ -190,6 +190,13 @@ namespace ShareX.HelpersLib
         NearestNeighbor
     }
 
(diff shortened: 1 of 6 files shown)


Jaex:
Thanks!

PR #5124, opened 2020-10-14
Fix for toolstrip overflow in image editor

nkosichek:
Minor bug I noticed when using region capture for small regions. Toolstrip in editor window overflows beyond displayed size of editor window. Not sure if this is mentioned by any issues.

This is how it looks without the fix
behaviourwithoutfix

This is how it looks with the fix
behaviourwithfix


Jaex:
It is already fixed. Because your repo is from 2019 you don't see the fix in your ShareX.

nkosichek:
13.2.1behaviour
Then I am confused. Is this behaviour exclusive to my machine?

nkosichek:
What's your resolution? I am 2560x1440.

Issue #4743, opened 2020-04-12
Color picker right click for RGB


JapanYoshi:
I'd like the option to copy the RGB integer values from the color picker *alongside* the hex code.

Aholicknight:
@JapanYoshi Already does that. 
![image](https://user-images.githubusercontent.com/7843719/79183305-bae0f600-7dd6-11ea-96a2-6d7cefc57f36.png)

JapanYoshi:
I meant the “Screen color picker”. As it stands, there is only one custom defined format available in the settings.

Jaex:
RGB is here: ![](https://jaex.getsharex.com/2020/04/ShareX_bcjtIU0pOp.png)

JapanYoshi:
No, I mean I want multiple formats to copy on a case by case basis. Right click does exactly the same thing as left click right now; why not default left click to hex and right click to decimal RGB?

Jaex:
Because right click let you not copy anything to clipboard.

JapanYoshi:
What about shift click then?

AfricansCantSwim:
i agree this si a good idea

tomastalian3:
![image](https://user-images.githubusercontent.com/55493945/95573819-5244a000-0a2c-11eb-8d5b-30bdfc2666d3.png)
Would something like this be okay?

JapanYoshi:
Sounds great!

Jaex:
Added in #5126

PR #5126, opened 2020-10-15
Added alternative screen color picker format.

tomastalian3:
Added entry for alternative screen color picker format to Task Settings -> Tools. As per request #4743 .
If holding SHIFT when you copy color info in screen color picker, alternative format is copied instead.

image


Jaex:
Why shift click instead of ctrl click? Isn't ctrl modifier used more than shift?

tomastalian3:
Not sure which one is used more. All I know is that shift and ctrl are both being used together with mouse click in various applications. But I can change it to ctrl if you want.

Jaex:
Thanks.

PR #5155, opened 2020-10-26
add Japanese translations

kanaxx:
I added some words of translation into Japanese which is based  13.2.2 dev version.


Jaex:
Thanks.

PR #5287, opened 2020-12-22
Fixed URL encoding for Plik

ZerGo0:
This adds encoding to the result URL so that it can be pasted in chats and so on without having to replace space with %20 and so on.


Jaex:
I think only file name part must be encoded so you can do:

```cs
result.URL = $"{Settings.URL}/file/{metaData.id}/{actFile.id}/{URLHelpers.URLEncode(actFile.fileName)}";
```

ZerGo0:
> I think only file name part must be encoded so you can do:
> 
> ```cs
> result.URL = $"{Settings.URL}/file/{metaData.id}/{actFile.id}/{URLHelpers.URLEncode(actFile.fileName)}";
> ```

Sure, changed.

PR #5288, opened 2020-12-23
Named pipe possible fix

OMGasm:
I don't know how to fully test it, but so far this seems to fix the problems fadc744799514cc8b44f406af038c724875906bf introduced.


OMGasm:
I should add, those problems being that a) shell actions and whatnot being unable to send their request to the main instance, and b) those instances never exited

Issue #1201, opened 2015-12-20
add Exit option to main window


smsrkr:
Can you add an Exit option to the main window? Preferably at the bottom of the left sidebar, after Donate and About options. Then ShareX can be closed from both the tray and the main window.

I know the tray icon can be disabled, and consequentially ShareX will exit when the window is closed, but then I'll miss the quick actions in tray.

Jaex:
I don't wanna increase main window height for unnecessary thing because some people use very low resolution and main window couldn't fit their screen then.

smsrkr:
well, how about an option like "Show Exit button in main window" in **Application Settings > General** ?

Jaex:
@jonfre if you disable tray icon from application settings then it works exactly like that.

StingyJack:
@Jaex - I added a [PR](https://github.com/ShareX/ShareX/pull/5331) to make the text align with most other programs that have similar functionality ("Minimize to Tray"). 

I can only do en-US as its the only language I know well enough. I hope thats not a problem.

PR #5331, opened 2021-01-17
#1201  - changing the setting text label to match other programs

StingyJack:


StingyJack:
Get deprogrammed @github-actions bot

StingyJack:
The @github-actions "stale" bot is most likely the result of a mixin between a ColecoVision and a TI-80, including the dementia from being so ancient (thats where it gets the gall to call anything else "stale").

Issue #5339, opened 2021-01-18
White background while in clipboard


Hxnry890:
Hello,

In ShareX, I always have a white background when I paste an image from the clipboard. I don't know if the error is on my computer side. Unfortunately, I have nothing in the clipboard settings. My operating system is Windows 10 Home. Please help, the white background in my screenshots doesn't look good.

Jaex:
![](https://jaex.getsharex.com/2021/01/ShareX_lWzrCZ4qU5.png)

mchubby:
Hello @Jaex 

I'm not sure if the question was spelled correctly, but for me the issue is how in image editor `Ctrl+V` paste the image that is already in the clipboard is rendered with a **solid color background**.

I have looked in the source code and do not understand why it behaves differently from adding a shape from a file since they both rely on the same class.

```
ShapeManager.cs

ShapeManager PasteFromClipboard()
|-  ClipboardHelpers.GetImage()  // bitmap with Format32bppRgb
|-  InsertImage()
    |- CreateShape(ShapeType.DrawingImage)  // new ImageFileDrawingShape
    |- shape.SetImage()
```

versus
```
ShapeManager  InsertImageFile()
|-  ImageHelpers.LoadImage()  // bitmap with Format32bppRgb
|-  InsertImage()
    |- CreateShape(ShapeType.DrawingImage)  // new ImageFileDrawingShape
    |- shape.SetImage()
```

Ctrl+V in Paint.NET for instance renders the pasted clipboard image with a transparent background.

A difference is the ImageInterpolationMode but I doubt it factors in. Another one is the bitmap's raw format which is a different guid, but does that matter?

PR #5438, opened 2021-03-03
fix: Handle DIBV5 when pasting to avoid a solid background

mchubby:
See issue #5339


Jaex:
Hi, I implemented it as optional setting using already exists functions in ShareX: https://github.com/ShareX/ShareX/commit/f060c8b7f4ff32e1481b2822979e529f48da4996

PR #5500, opened 2021-04-01
Add Vault.ooo uploader service

shjonas:
Adds Vault.ooo service that works with screenshots and large files. Files are encrypted before upload and decrypted in the browser. Usage does not require an account.


Jaex:
Why you also added it to image destinations list?

shjonas:
Vault.ooo is not a pure file host, because certain file formats (such as png) are displayed directly online without download. Most file hosters will download all files, whether they are images or not. To let users know that this service is also designed for sharing screenshots, I added it in addition to the image destinations. I think this way it is clearest for your users, but you can remove vault.ooo from the image destinations if you need to.

Jaex:
There is a lot of file uploaders in ShareX which can show image preview too but I did not add them to image uploaders list because then it would be duplicate entry to file uploaders sub menu.

shjonas:
No problem, I just thought it was more understandable that way. I personally don't mind duplicate entries in different categories. However, feel free to remove the entry from the image destinations :)

shjonas:
I think most users don't use the file hoster for images because they assume that the screenshots have to be downloaded without preview. Any service that has a preview is an image hoster for me and should therefore show up in the image destinations list. But I also see the problem with duplicates and would understand if you just add the Vault.ooo entry to the file destinations only.

shjonas:
Hello, I haven't heard from you for a while now, so I wanted to ask if you have checked my commit yet. Are there any changes needed from my side or is everything ok?  You seem to be bothered by the double entry in the file uploaders and image hosts. Should I remove this duplicate entry or do you want to make it so that it suits you? Can you tell me approximately how long it will take you to check and merge my commit?
Thank you in advance!

shjonas:
Hi, since I originally added vault.ooo to both image uploaders and file hosts and it was bothering you, I've now changed that. Vault.ooo is now only in the file hoster section. Please let me know if everything is okay now!

Jaex:
> Hi, since I originally added vault.ooo to both image uploaders and file hosts and it was bothering you, I've now changed that. Vault.ooo is now only in the file hoster section. Please let me know if everything is okay now!

Thanks.

PR #5514, opened 2021-04-09
Make string building effective

Banyc:


ShareX.HelpersLib/CLI/CLICommand.cs
```diff
@@ -1,5 +1,7 @@
 ﻿#region License Information (GPL v3)
 
+using System.Net.Mime;
```
Banyc, line 3: This line is auto-generated. I forgot to remove it.

Issue #5501, opened 2021-04-02
Be able to use Roman numerals in the step annotator


KeronCyst:
ShareX 13.4 on Windows 10 Home 20H2:

1. Start taking a screenshot
2. Press "i" to enter the Step annotator
3. Check the `Shape options`

Expected result: Be able to choose between lowercase or capital Roman numerals, along with the current options
Current result: There is no such ability

PR #5519, opened 2021-04-16
#5501 added ability to use roman numerals in the step annotator.

EdwardSierra:


Jaex:
This is not good idea having two bools:



It should be list. So in future more step types can be added to enum.

Jaex:
Thanks!

PR #5548, opened 2021-05-03
Drag drop notifications with threshold

Craftplacer:
Similar to the original PR, this adds drag drop support to notifications with filenames.

This time the NotificationForm is tracking how far the mouse is moving until it triggers DoDragDrop(...) to prevent click-hijacking.


Jaex:
Tested now, it seams working without problem.

PR #5564, opened 2021-05-09
Add thumbnails to videos in task previewer

Kampfkarren:
Before:

ShareX_2021-05-09T05-09-42

After:

image

### DIFF
--- a/ShareX.MediaLib/VideoThumbnailer.cs
+++ b/ShareX.MediaLib/VideoThumbnailer.cs
@@ -29,6 +29,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Drawing;
 using System.IO;
+using System.Threading.Tasks;
 
 namespace ShareX.MediaLib
 {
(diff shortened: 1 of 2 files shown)


Jaex:
Using FFmpeg executable is not good idea because it is slow to run executable and not everyone have FFmpeg installed in their system. Instead need to use API which Windows Explorer using when showing video thumbnails, but I couldn't find how to do it.

Kampfkarren:
Is this what you want? https://stackoverflow.com/a/60425444/1910744

Jaex:
I prefer to not add another NuGet package, solution must be similar what I'm using to get file icons currently.

Kampfkarren:
I did some diving, and it seems to just boil down to [IShellItemImageFactory::GetImage](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ishellitemimagefactory-getimage).

dekdevy:
An alternative way to handle this would be to store video thumbnails as images in a temp / thumb folder somewhere. optionally, of course. this would make sharex significantly smoother to use for many people im certain.

Issue #5432, opened 2021-02-27
ShareX sometimes does not capture HDR screen contents


PredragBoksic:
Often it works well on my Windows 10 with Nvidia GTX 970. Sometimes, the captured images look pale whitish with insufficient details. That was the default a long time ago when the HDR was unsupported. Now it happens rarely.

PredragBoksic:
I have some new observations. More precisely, the occasional inability to capture accurate saturation and contrast affects only the surface of the Chrome browser. The captured screenshot is sometimes very bright and poorly visible as if the HDR 10 Windows setting is set to 100% instead of 50% (default). The Windows setting deals with the SDR/HDR proportion. 100% HDR setting is no good. The image in all programs becomes overexposed. So, the Chrome document window contents mysteriously end up looking just like that in the eye of ShareX.

Issue #4570, opened 2020-01-13
HDR in Windows 10 on GTX 970 looks pale and unusable


PredragBoksic:
Screenshots appear pale and unusable in HDR 10 mode (that is used in combination with the good old LG Ultrawide monitor with good results).

sylveon:
Duplicate of #3751 and #4105

Issue #4105, opened 2019-04-29
Windows 10 UHD mode completely washes out visuals when taking a screenshot


StormedJF:
I recently got a UHD 4K TV and noticed after enabling Windows 10's HDR mode (also known as Windows HD Color), all visuals are be washed out when entering screenshot mode. As an example, here is a comparison:

Original Image: https://i.imgur.com/Jh7RhPj.jpg
Screenshotted Image: https://i.imgur.com/f0LFTc9.jpg

While I don't know if this would even be able to be fixed since this is probably Windows fault with how it handles HDR, but I feel like it's worth sharing here.

dylanh724:
I have this exact issue, and only happened when I enabled HDR10 on Win10.

PR #5565, opened 2021-05-10
[WIP] still-image HDR capture support

imbushuo:
This (draft) PR implements HDR support discussed in issue #3751, #4105, #4570 and #5432. On the high-level overview, it bypasses the old GDI+ capture pipeline using Windows Runtime Graphics Capture API (a wrapper of DXGI Desktop Duplication) while providing a GDI+ compatible 8-bit SDR image that tonemapped from HDR capture source, which covers the majority usage scenarios. A few open questions remain:

# Open questions to be discussed & solved
- Need to choose between reducing overall luminance or clipping while processing the SDR output. The currently implementation reduces overall luminance for final output, similar to what the Xbox Game Bar capture does. On other platforms such as macOS, they decided to clip regions with very high luminance to keep SDR content's brightness. Or alternately, use HLG (?) 
- Which tonemapping algorithm to use? Or let user choose their favorite one? The current PR implementation uses Reinhard, which works well for many test scenarios. But there a few more to consider, like ACES (need to have a D65 white point version though), simple clipping (but something will be washed out) and more.
- Do we need to save the original HDR capture? Not useful for sharing but ideal for post processing out of ShareX.

# Known Caveats
- It doesn't work with scrolling capture, GIF and video capture now. Ideally if video capture switch to this new capture pipeline, it can leverage GPU processing power instead of using ffmpeg.
- A bit more memory usage. Stable 220-400 MB memory usage on a 192GB memory computer with two 4K HDR monitors.

# Sample Captures
msedge_HOgcv742iw
ApplicationFrameHost_LgtASlrQ9z
ApplicationFrameHost_DtfpurTfia
Acbaep2Sqq


craigkc:
Any chance you're still working on this?  I looked over the code and it appears clean and straight-forward, but it's unlikely I have time to contribute outside of testing.  Without this, I may have to abandon using ShareX, and it seems clear many others are facing the same issue.

Let me know how I can help.

imbushuo:
I think I still have a few corner cases to deal with, and the main author has some concerns regarding introducing external dependency (ShareDX is a lot.) I am currently working on this but the bandwidth is not currently guaranteed.


Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

craigkc:
Thanks for the reply.  With SharpDX having halted support some time ago, I can understand the issue there.  I'll take a better look at the rest of the ShareX code to see the prevailing style/intent and see if I can help.  I don't want to pollute your PR like an issue discussion, so I'll leave it at that for now.

rmarskell:
I would like to see this issue re-opened as there are currently no open HDR issues and this issue seems to be the most promising. This is still a big problem for those of us with HDR enabled.

rikai:
This is something that has been a personal annoyance to me, as it regularly ruins my screenshots on any HDR-enabled system that has HDR enabled.

I did see that there was hesitation to use SharpXD because it had been continued, however it does look like there are a couple of alternatives available that seem to have at least some level of HDR-awareness/support:

[Silk.NET](https://github.com/dotnet/Silk.NET) (official .net foundation project)
[Vortice](https://github.com/amerkoleci/Vortice.Windows) - independent, but actively maintained

Unsure if these projects are actually suitable for the use case, but I thought I might mention them. :)

WDD3MS:
I have the same problem with HDR. (LG monitor + AMD 5600G (Vega7 ) via DP]
Usually the captured image is heavily overexposed.

pish180:
This is marked as closed but it still an issue.  Any reason this wasn't merged?

SalmonSays:
This really should get merged.

WDD3MS:
I notice this only on EDGE.

ividyon:
Any news? This is vital for anyone trying to use HDR displays.

WDD3MS:
I have still this problem. 
Only on EDGE, and some web is okay, some don't. 
Even on the same page, and on the same feed (live updated) some news part with picture - overexposed, some - ok.

diff-by-default:
+1

This would be a quality of life gamechanger for almost everything I do at my computer these days. HDR is everywhere and this issue is only getting worse with windows 11 having auto HDR.

spartaque12:
its been several years, still no fix, amazing

rmarskell:
> its been several years, still no fix, amazing

Yeah, the dev has said it doesn't affect them so they're not interested in fixing it. 🤷

Also, I've tried running the branch for this PR and it didn't seem to work for me. I may have been doing something wrong though. Not 100% sure.

MRDGH2821:
Any chance this is still worked on?

ScottRFrost:
Still a problem in the 16.x.x series

Psyda:
Still a problem. Any ideas? 
QfHoRHY 1

ScottRFrost:
For anyone finding this (after years) via search / etc, as a temporary workaround, you can do a HDR friendly screen capture on Windows 11 by Press Windows Key + Alt + Printscreen, which will save the file by default to C:\Users\YourUsernameHere\Videos\Captures

You can then load this image into any normal photo editing app (such as paint.net) to mark it up before sending

Hopefully we won't need this workaround once ShareX supports HDR natively!

PR #5663, opened 2021-06-27
localize border style

kanaxx:
I added some codes to localize "Border Style" enum text.
Could  you please check and merge them?


Jaex:
Thanks!

Issue #3888, opened 2019-01-17
"Last vertical" adjustment only works with the last image.


hizamakura:
In Scrolling Capture, the "Last Vertical" adjustment option will not work unless it is the very last image, even if you changed the amount of images by using the "Remove Last" option. It should adjust the current last image when you try to remove some images, as how it implies.
![Image](https://thighs.are-pretty.sexy/add2e3.gif)

mateherber:
This still happens in latest version

supra107:
Still happens in 13.7, meaning that when you inevitably take too many screenshots at the end you cannot properly adjust the final vertical alignment as it's being applied to an unwanted image that's been excluded by the "Remove Last" option. It should be reopened as it essentially forces users to do the final alignment manually in other software.

rakosi2:
@mateherber Could you reopen this as I also am having this issue with 14.1

PR #5690, opened 2021-07-07
Update ScrollingCaptureForm.cs

mateherber:
Fix for #3888


Jaex:
There is reason it is not using last image for it because only last scroll have different size so if you remove last images then it not becomes last scroll so it using normal vertical option for it on purpose. So how it is fix? You can break some peoples workflow like this.

mateherber:
It should not break anybody's flow as currently if "Remove last" is greater than 0 value of "Last vertical" is never used so potentially left to 0 too. In my use-case when auto-stop is selected the last two images are seemingly same (must be some minor difference as it is not removed by the "remove duplicate images" option. So essentially I cannot achieve what I want.

mateherber:
Hm that is true but what I experience is slightly different. Not just the last image has different offset but the previous too (like it had  slowed down. I'll prepare a image to showcase

PR #5709, opened 2021-07-20
Fix cursor offset in DrawCursor method

Jomtek:
When I tried to take a screenshot - indeed making sure the cursor was set to "visible" in the settings - I could not see my cursor in window capture mode or monitor capture mode.
I figured out the problem was coming from the CursorData.DrawCursor method, which was applying another unnecessary offset by using CaptureHelpers.ScreenToClient.

Just removing line 115 actually solved the problem, and now I can see my cursor 👍

### DIFF
--- a/ShareX.HelpersLib/CursorData.cs
+++ b/ShareX.HelpersLib/CursorData.cs
@@ -111,9 +111,10 @@ namespace ShareX.HelpersLib
         {
             if (IsVisible)
             {
+                // Apply offset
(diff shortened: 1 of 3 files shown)


oparkerj:
There are certain places where the offset is necessary. [This commit](https://github.com/oparkerj/ShareX/commit/940afc8a27d264c497f9efe84113a71de229446e) properly fixes the issue, keeping the offset when needed and disabling it when unnecessary.

PR #5761, opened 2021-08-13
Fix typo in english.iss

eltociear:
occured -> occurred


Jaex:
We updated InnoSetup script now and those language files are not used anymore.

PR #5770, opened 2021-08-19
Localize the Type summary in History window title

L1Q:
- Create ResX entries for EDataType ENUM
- Implement a lookup dictionary: enum string value => translation string
- Query the lookup with strings found in history file

Thanks OMGasm


Jaex:
Thanks.

PR #5791, opened 2021-08-27
Added strength parameter for some image filters

evancg37:
Added a strength parameter for the Sharpen, Smooth, and Mean Removal image filters. 

### Details

The 3 affected filters use image manipulation methods from the ShareX.HelpersLib.ConvolutionMatrixManager class. These methods already accept a "weight" integer as an argument that determines the strength of the effect, so this change just adds a parameter in the GUI that sets the value of that "weight" argument.

The Strength integer is capped between 1 and 10 for all three affected filters, and the default is 5. The strategy used for adding the parameter is identical to how it is for other filters with similar parameters, like the Blur filter's "Radius" setting. The integer value is not directly used as the argument to the image manipulation methods - instead we apply strength as an offset from a constant arbitrary value, which I tuned for each filter so that the user perceives a comfortable relative strength scale from 1-10. 

The weird part about this code change might be the arithmetic and constant arbitrary values used for the weight argument calculation. The relationship of the ConvolutionMatrixManager "weight" to perceived strength of the effect wasn't linear, so I can really only say that I tuned it the best I can. 

### Purpose
I have been using the Scale effect in combination with the Sharpen effect to compress my screen captures while keeping them sharp. However, the Sharpen filter was too strong and makes my screenshots look bad and unusable. I wanted to reduce the strength of the Sharpen filter, but there was no option for that.

### Testing
All 3 filters still function correctly after adding the change. The strength parameter now shows up in each affected image effect inside the Image Effects editor window. 

For my own use case, I set my sharpen filter to 2/10 just to apply a slight sharpen after I scaled down my images. Works great!

### DIFF
--- a/ShareX.ImageEffectsLib/Filters/MeanRemoval.cs
+++ b/ShareX.ImageEffectsLib/Filters/MeanRemoval.cs
@@ -32,11 +32,32 @@ namespace ShareX.ImageEffectsLib
     [Description("Mean removal")]
     internal class MeanRemoval : ImageEffect
     {
+        private int strength;
+
+        [DefaultValue(5)]
+        public int Strength
+        {
+            get
+            {
+                return strength;
+            }
+            set
+            {
+                int capped = value.Min(10);
+                strength = capped.Max(1);
+            }
+        }
+
+        public MeanRemoval()
+        {
+            this.ApplyDefaultPropertyValues();
+        }
+
         public override Bitmap Apply(Bitmap bmp)
         {
             using (bmp)
             {
-                return ConvolutionMatrixManager.MeanRemoval().Apply(bmp);
+                return ConvolutionMatrixManager.MeanRemoval(30 - (2 * Strength)).Apply(bmp);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


Jaex:
I'm not fan of the arbitrary values you used. Also why you added description to those classes?

evancg37:
I understand if it's not a good idea to do it that way. Description being added was just a result of copying the ideas from other classes.

boehs:
> I'm not fan of the arbitrary values you used.

Are you talking about `[DefaultValue(n)]` or `ConvolutionMatrixManager.Filter(n - (2 * Strength)`?

Jaex:
I'm talking about `ConvolutionMatrixManager.Filter(n - (2 * Strength)`.

evancg37:
An alternative could be to directly use Strength as the filter argument like ConvolutionMatrixManager.Filter(Strength) so that no arbitrary calculation is necessary. The filters just behave weirdly with Strength values around 0, so a minimum could be in place or something.

boehs:
https://drewdevault.com/2021/10/26/stalebot.html

PR #5808, opened 2021-09-08
Improve HistoryForm file type dropdown

L1Q:
- Make dropdown options localizable
- Fix the dropdown options not updating with history refreshes


Jaex:
Thanks.

Issue #2861, opened 2017-10-18
Disable upload via registry/Group Policy


mcmcghee:
Is there a registry key that can be set to disable uploads? 

This would be needed for deployment in certain environments to prevent accidental upload of sensitive data.

Jaex:
Only this: ![](https://jaex.getsharex.com/2017/10/ShareX_zBlBf6B5V6.png)

tekmaven:
Hey @Jaex is there a way to get this set without changing it in the UI?  I think the goal here for most requests around DisableUpload is that we would like to use ShareX in a corporate environment, and this seems to be a real barrier for administrators.

Jaex:
Even if you change it outside UI still user can re enable it from settings.

tekmaven:
I have a custom fork where I've modified the `DisableUpload` property to always `return true;`.  I think we would want the registry setting (or some other means of configuration) to never allow that boolean to be false.

tekmaven:
@Jaex I believe ShareX could have a lot more adoption in businesses if there was a administrator-controlled way to enforce screenshots do not get uploaded anywhere.  The default behavior of public uploads with a confirmation screen on the first upload does not work in a corporate environment and probably scares away a lot of administrators.

Would you be willing to accept a PR that changes this behavior?

Abhishek-k2020:
did we have any settings through group policy / firewall , which can enable us to restrict uploads on public code repositories

Jaex:
This feature added in #5822, let us know if implementation fits to your needs or requires any changes before release.

PR #5822, opened 2021-09-14
Added registry check to allow system admins to disable uploads

Erik3003:
I added the ability for system admins to disable uploads system wide by setting a registry key.
This is especially useful in corporate enviroments for work with confidential files and has been requested before in issue #2861
The registry key is: "HKEY_LOCAL_MACHINE\SOFTWARE\ShareX" and an entry needs to be "UploadForbidden" with the value a of "true" and needs to be created manually if a user wishes to use this feature.

Also added a missing German translation for "ThisFeatureWillNotWorkWhenDisableUploadOptionIsEnabled".

### DIFF
--- a/ShareX/ApplicationConfig.cs
+++ b/ShareX/ApplicationConfig.cs
@@ -319,11 +319,5 @@ namespace ShareX
         public List<Color> RecentColors = new List<Color>();
 
         #endregion Color Picker Form
-
-        #region Formless
-
-        public bool UploadForbidden = false;
-
-        #endregion Formless
     }
 }
--- a/ShareX/ApplicationConfig.cs
+++ b/ShareX/ApplicationConfig.cs
@@ -319,5 +319,11 @@ namespace ShareX
         public List<Color> RecentColors = new List<Color>();
 
         #endregion Color Picker Form
+
+        #region Formless
+
+        public bool UploadForbidden = false;
+
+        #endregion Formless
     }
 }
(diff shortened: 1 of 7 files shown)


Jaex:
I'm not sure is it good idea to check for registry each upload, wouldn't it be better idea to check it once at startup of the app? Especially because most people not gonna use this system option, therefore each upload it will be doing unnecessary checks for most users.

Jaex:
Also I have concerns about people gonna require `HKEY_CURRENT_USER\SOFTWARE\ShareX` path to be used instead, so they can have user specific control on it.

Erik3003:
I changed it to only check the registry on startup and have settings from HKEY_CURRENT_USER overwrite HKEY_LOCAL_MACHINE. Furthermore, I changed the entry to "UploadForbidden", to be consistent with the new variable naming.

Jaex:
You can store global variable in Program.cs instead of ApplicationConfig.cs because that file is used for settings which is saved to file. So it is not safe to store readonly global variables there.

Jaex:
Thanks!

Jaex:
Please note I changed registry name back to "DisableUpload".

BoBiene:
@Jaex any ETA / plans for the next release containing this feature?

Jaex:
There is no ETA.

@Erik3003 I think "LocalMachine" should have priority over "CurrentUser". For example if system admin set HKLM key and if user have access to HKCU then user can override system admins flag. So only if HKLM key is not exists then ShareX should check for HKCU key. Currently it is opposite.

sylveon:
The intent was probably to use in an environment where registry editor access was disabled to users, but yeah better do it that way (since they could use third party to write to the registry still)

Issue #674, opened 2015-05-09
Request to support Microsoft OneNote


CleverClogs:
Hi Jaez and Team,

Have you considered adding support for Microsoft OneNote to the ShareX output options? Should you be interested: the developer platform is at http://dev.onenote.com/.

Thank you very much,

Marjolein

Jaex:
Who is Jaez.

CleverClogs:
Oops, apologies for that, Jaex. 
Microsoft OneNote is a note-taking program. It's used a lot in education, but also by knowledge workers, information professionals, lawyers, writers, bloggers.
Thank you kindly for considering it.

Jaex:
So how ShareX can have output of it?

CleverClogs:
After a user creates a screen capture, ShareX prompts the user and asks in which OneNote notebook and specific page the capture should be stored. Several other screen-capturing tools have implemented it over recent years. Does the documentation on their API pages give you enough clues?

creeperparty568:
OneNote is supported.

CleverClogs:
Can you give me a hint where in ShareX I can find the OneNote output option?
Thank you!

unnes:
I don't mean to necro, but there may be some confusion about OneNote -- it's not the same as OneDrive.

OneNote is a collaborative note taking application akin to evernote.

Adondriel:
Actually, OneNote has a really bad, ugly crappy screen clipper that is a pain in the ass to use, but it exists... sadly. Also, sorry for necroing this post... This would be a neat addition if it could work.

bweaver777:
Please add OneNote integration. It would be nice to have a ShareX screen shot go directly to a default section in OneNote, optionally with editing along the way, and optionally changing the target section in OneNote.

With the new people switching from Evernote to OneNote lately, it'd be a good time for the feature.

Jaex:
Isn't OneNote is paid software?

dannoe:
Nope, it's completely free.
https://www.onenote.com/

bw777:
Free. There are a couple "editions": the Windows App Store version, which has a few unique features but is generally considered to be inferior to the Desktop version. Most people who use ON seriously use the desktop software.

http://www.onenote.com/download

Jasius:
OneNote support would be great.

Adondriel:
Yea, the basic version is garbage. My tablet came pre-loaded with it and it was trash compared to the full version.

alsab200:
[...]

PR #5863, opened 2021-10-04
Add Default Printer Override

tajetaje:
Add a way for users to, if `Don't show Windows print dialog` is enabled, set a custom default printer for ShareX to use when printing.

This is mostly a quick and easy way to solve #674 but obviously also has a number of other applications. I tried to stick to the existing code style as much as possible.

### DIFF
--- a/ShareX.HelpersLib/Printer/PrintHelper.cs
+++ b/ShareX.HelpersLib/Printer/PrintHelper.cs
@@ -94,6 +94,26 @@ namespace ShareX.HelpersLib
             }
         }
 
+
+        public void TryDefaultPrinterOverride()
+        {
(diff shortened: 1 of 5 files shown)


Jaex:
Your pr changed location and size of each controls in `ShareX/Forms/ApplicationSettingsForm.resx ` file.

tajetaje:
Yup, you are right, it seems like Visual Studio saw fit to add a margin to each and every control that I loaded into the form editor. That should have fixed it, seems like everything is still working fine. I just copied the controls I added onto the original file manually without loading it in Visual Studio.

Jaex:
Thanks.

PR #5900, opened 2021-10-22
Add translator site dropdown for OCR window

IAmVisco:
### Summary
Added a dropdown in settings to select default translator website that will be opened if you click the link in the OCR pop-up window. Can be easily extended to add more sites like Yandex, Bing etc. later.

### Not-so-sure points:
- Used pseudo-templating to replace site name in the link - problems with l10n (was this solved before?)
- Couldn't find how to add loc line for the new label except adding it one by one into each .resx file (is there a better way?)

Haven't coded on Windows Forms for ages but hopefully it looks okay-ish.


IAmVisco:
Took more then expected. Played around a bit when I had time and decided to stop with enum.

Jaex:
I prefer to not add by default foreign (non english) services or less known services. Thats why it is best to let users configure whatever they want.

IAmVisco:
I thought we could close those 2 issues you mentioned earlier but sure. The rest should have some use as defaults.

Jaex:
Thanks.

Jaex:
> That's why it is best to let users configure whatever they want.

Added it now:

PR #5925, opened 2021-11-06
Added new image combine methods

DylanMcBean:
Added new Image Combine method

UI Changes
ShareX_V4z0BOFmJ6

Converts from things like this
wiAqBAMmXq

Into things like this
v0hQ4eLiU9

Pros:
* can help layout images in a nicer to view style without the user having to mess with combining multiple vertical/horizontal groups of images
* can help save on space as these should take up less pixel space and therefore less memory for the user
* much quicker than if the user was to do this by hand manually

Cons:
* as a first draft, there could be some optimizations added onto it.
* can create strange image sizes as it is not constrained to one axis.


Jay-Tarzan:
Web-site

DylanMcBean:
what do you mean by Bot net and Gut hub?

DylanMcBean (commented):
looks good

Issue #2325, opened 2017-02-27
Request: Add zoom function to image editor


Freekers:
Currently, the image editor does not have a zoom function, which makes it hard to edit small images or precisely crop etc. Thank you!

edphemeria:
+1, I sometimes annotate very large files which require lots of scrolling.
Huge thanks for ShareX, it's just the best!

Davexx:
Yes, I have 2 monitors and sometimes I cannot access portions of the screen capture.  I would love this feature as well.  and thanks for the awesome tool.  It's very thorough!

nepomuc:
Yes, please add this feature! The image editor is so feature-rich, it even has two different ways to anonymize regions by blurring or pixelating them. But it has no zoom in/out.

TJohn7002:
Vote for this

Freekers:
1 year later and I'm still rooting for this feature ;) Pinging @Jaex

marcoscheel:
+1 for this feature

FELITH:
me too

markusd1984:
👍 +1 I discovered this because the picture I opened was too large. 
Without a zoom function, one can't view the image as a whole and to edit needs to scroll around :)

viper25:
Yes, this is one basic feature missing from such an advanced snapshot tool. I was using CTRL + mouse scroll  to zoom in/out instinctively and was mildly surprised to see one couldn't zoom.

Jaex:
Adding zoom is not basic at all.

Davexx:
@Jaex  I believe @viper25 meant "fundamental" when he used the work "basic".  I don't think he was diminishing the effort it would take to add the code.  It's one of those things you quite literally expect to be there especially with such large screen capture real estate these days.

Jaex:
I don't think there is any software with on screen annotation which can zoom. It is not image editor, it is region capture with annotation support.

nepomuc:
@Jaex Do you mean something like this?

![screen_annotation-do_you_mean_something_like_this](https://user-images.githubusercontent.com/864660/40316689-64a37120-5d1f-11e8-8ef3-d008c4306d56.gif)

Jaex:
That is image editor, not on screen annotation. Their region capture don't have zoom and annotation functionality.

nepomuc:
[...]

PR #6084, opened 2022-02-08
Support zooming in image editor

EricTetz:
- zoom with CTRL+mouse wheel (zoom centered on mouse)
- zoom with CTRL++ or CTRL+- (zoom centered on window center)
- reset zoom with CTRL+0
- zoom to fit with CTRL+ALT+0
- "Zoom to fit on open" option
- zooms by 1% of current size per wheel tick, so it feels consistent regardless of zoom level
- zoom range is 20%-600%

Implementation is to apply a scale transform when rendering. Some objects need to ignore this transform when rendering (magnifier, scrollbars, button tooltips, etc.) and some operations that use mouse coordinates need scaled coordinates to work correctly. 

This implementation scales up the annotations tools, like control nodes, along with everything else. They can get unusably large, which is the only reason there's an upper limit on zoom. Addressing this (if it was deemed worth it, as it increases code complexity) would have to be a further effort.

ZoomDemo


derekantrican:
This is wonderful. Thank you so much for taking this on.

A couple comments: 

- Suggest removing the "ZOOM: ##%" text from titlebar when zoom is at 100%
- Suggest some key combo to reset zoom to 100% (not sure what key combo - `Ctrl+0` is what is used in most browsers & Electron-based apps)

emilio-cea:
So I understand this will be coming in future releases? ❤

Jaex:
Thanks for all the work @EricTetz!

I'm testing it right now, currently I noticed that zoom also works in region capture and causing image to shift position.

But most importantly, even without zoom still FPS of region capture is lowered. Probably have to add plenty of `ZoomScale != 1` checks to code to make sure that FPS is not lowered when there is no zoom at all.

L1Q:
Great effort! Already this feels completely usable for every person so eagerly demanding the feature.

Here are all the issues I've encountered while testing:
- Control handles are scaled with the image, especially hilarious at 400+% zoom
- Modifying canvas (crop, rotate 90%, flip) while zoomed misplaces the image on the screen
- Vector-based drawings (arrow, line, freehand, etc.) are rendered in screen space, not in image zoom
- No keyboard hotkey for zooming in/out. Something like ``Ctrl``+``-`` and ``Ctrl``+``=`` would be a great help.

What I could see during editing:
image
What I get after the saving:
image

But that's minor gripes IMO.

EricTetz:
@Jaex 

> I noticed that zoom also works in region capture and causing image to shift position.

That's just the pan portion of the zoom not being disabled during capture. Fixed.

> even without zoom still FPS of region capture is lowered

I don't have any difference between no ScaleTransform and ScaleTransform(1,1) on my machine. How are you testing?

I just checked in a change that disables any calls to ScaleTransform if zoom is 100%. See if that fixes it for you.

EricTetz:
@L1Q 

> Control handles are scaled with the image, especially hilarious at 400+% zoom

Doing it differently is a much bigger edit. Zooming everything is how Greenshot (another open source capture tool with an editor) does it, too, presumably for the same reason. For the first pass it would be a lot easier to just limit the max zoom factor.

> Vector-based drawings (arrow, line, freehand, etc.) are rendered in screen space, not in image zoom

Not sure what this means.

> No keyboard hotkey for zooming in/out. Something like Ctrl+- and Ctrl+= would be a great help.

Easy enough.

> Modifying canvas (crop, rotate 90%, flip) while zoomed misplaces the image on the screen

Ah, some centering routine probably needs updating. Don't have time to look right now.

Jaex:
> I don't have any difference between no ScaleTransform and ScaleTransform(1,1) on my machine. How are you testing?
>
> I just checked in a change that disables any calls to ScaleTransform if zoom is 100%. See if that fixes it for you.

I test by opening region capture and checking fps there.

Tested it again now. 147 fps in develop branch, 127 fps in your branch. I tested both your previous commit and last commit, fps not changed at all, it was 127 fps before and after.

Because there is no zoom support in region capture, fps should be staying same instead of dropping 20.

EricTetz:
@Jaex At some point in my initial hack, I removed the code that sets the compositing mode to SourceCopy when drawing the background. Performance was lost to alpha blending. Restoring that line recovers the missing performance when not zooming. Just checked it in.

Jaex:
I tested your latest commit now, FPS is fine when zoom is at 100%.

> Modifying canvas (crop, rotate 90%, flip) while zoomed misplaces the image on the screen

So only this centering issue is left.

derekantrican:
Fixes #2325

EricTetz:
Fixed centering.

I added **CTRL -** and **CTRL +** as hotkeys for zooming in/out. The zoom center for these operations is the center of the client area, which is how Photoshop does it.

When you zoom with the mouse wheel, the zoom center is the mouse position.

Jaex:
I noticed there is panning issue when zoomed in:

https://jaex.getsharex.com/2022/02/ShareX_7LvJXg5XSm.mp4

So when you try to pan slowly, it never pan, unless you move cursor fast.

Jaex:
I also noticed that scroll bars not accounting zooming correctly:

EricTetz:
> So when you try to pan slowly, it never pan, unless you move cursor fast.

I notice this. It's a straightforward problem, but requires changing a lot of code to fix.

When we're zoomed, panning movements have to be scaled. If we're at 6x zoom, we apply a 6x scale transform when rendering. To move something 1 pixel to the right with that scale transform, we *internally* move it 1/6 to the right. That's ~0.17, a floating point value. That's a non-issue with a drawing API. They all support floating point.
              
However, ShareX internally uses int for everything. You can't move something by 0.17. So if you move the mouse slowly, you're getting deltas of 1 or 2 that get rounded off to 0 when scaled and nothing moves. If you move the mouse quickly, the mouse delta gets large enough that the scaled value isn't rounded to 0, and it moves.

The fix is to just use floating point internally.  Has no effect on performance. It mostly means changing a lot of `int` to `float`, adding `F` suffix to some drawing types (e.g. PointF, RectangleF), and getting rid of superfluous truncating to int.

I tried this. I changed scaled mouse coordinates to float, which requires changing stuff ShapeManager.MoveAll, which means changing BaseShape.Move, so on and so forth, cascading out through the entire system. It's like changing const-correctness in C++ code. You can just change one bean without changing the whole burrito.

In any case, it works as I expected:

ZoomPan

This also fixes other things, like magnified drawing, which was being quantized:

ZoomDraw

However, I stopped short of committing this change, because it touches a lot of code:

[intToFloatDiff.txt](https://github.com/ShareX/ShareX/files/8057777/intToFloatDiff.txt)

So it really comes down to you, as the owner and caretaker of this repository. Are you willing to accept that big a change, or would you rather live with chunky panning when zoomed?

EricTetz:
> Vector drawings look smooth and beautiful as they are matching your screen resolution, but they do not match the image resolution. 

Right. We'd have to dynamically rasterize all the annotation tools, which is way beyond the scope of this change (mostly adding ScaleTransform(x) to the renderer).

> check my example screenshots posted earlier in the thread

Right. But to be fair, the labelling of the screenshots is misleading. What it should say is "what I see during editing" and "what I get after saving *if I then zoom in by 900%*". This is more accurate:

What I see during editing:

DuringEditing

What I get after saving:

AfterEditing

Which is much less jarring.

Jaex:
> The fix is to just use floating point internally.

Yea I guessed problem was to do with all coordinates using integer in ShareX region capture form.

> So it really comes down to you, as the owner and caretaker of this repository. Are you willing to accept that big a change, or would you rather live with chunky panning and quantized movement when zoomed?

If moving everything to floating point not causes any side effects while using the region capture at 100% zoom, such as one off pixel issues etc. then sure I don't see any reason not to use floats. I'm glad you did not push it to this PR, because making it separate PR would be better idea. That way I can check code differences more easily.

Thanks for the good work! After finalize the float changes, then we can ask users to test it via dev builds.

EricTetz:
> If moving everything to floating point not causes any side effects while using the region capture at 100% zoom, such as one off pixel issues etc.

Shouldn't effect that at all. We're just deferring the quantization to a concrete pixel until as late as possible. In fact, we don't even do the quantizing, System.Drawing does.

> making it separate PR would be better idea

That's going to be rough, given that I started with my zoom changes. I then basically did this: change Pan(int,int) to Pan(float,float), build, see a handful of errors. Fix a few of those errors, which is creating new errors, as we've changed code that's visible to a wider scope. Build. See a couple hundred errors now. Fix the errors on screen or in the active module. Build. Wash, rinse, repeat... *for a few hours*. Eventually the list is just 100 errors, some time later just 50, then 20, then *ding*, build successful. I ended up changing 450+ lines.

I'd rather not do that again, if at all possible. One alternative would be to stash the float changes, fix up any remaining bugs (like the scrollbar issue you noticed), then accept this pull request. We could even constrain the zoom level even more severely, to minimize the issue.  Then we *improve* zoom behavior by switching to floating point. Thoughts?

Jaex:
> EDIT: Just noticed the pull request says "merged". Does that mean you accepted the zoom stuff and I can do a separate pull request for the float stuff? If so, disregard what follows.

Yes pr is already merged.

Jaex:
I changed zoom levels to be like this now:

20, 30, 40, 50, 60, 70, 80, 90, 100, 125, 150, 175, 200, 250, 300, 350, 400, 450, 500, 550, 600 %

EricTetz:
> I changed zoom levels

Those feel good.

Side note: the big refactoring you did made the merge a nightmare. The floating point stuff touches a lot of code. If you could avoid cosmetic changes (like moving stuff to a different line in the file) for a couple of days, that would make the floating point stuff a lot easier.

dcarapic:
I've just cloned the repo and build the project. Zoom works great. Any idea when there will be a new release with this functionality?

PR #6087, opened 2022-02-09
Support FileExtension in Upload Path string regex

crscillitoe:
I'd like to be able to sort files on my s3 bucket by their file type. This PR adds support for a %fe (FileExtension) regex pattern to the "Upload Path" parameter.


crscillitoe:
As per discussion on Discord, I have renamed FileType --> FileExtension to more clearly indicate the purpose of this new pattern. I've updated the code aswell to behave less undesirably in scenarios where a file extension does not exist on the target file.

crscillitoe:
accidentally combined two branches, will re-open PR after cleaning up.

PR #6108, opened 2022-02-16
Switch to floating point internally. Panning fixes.

EricTetz:
Change a lot of internal code to floating point so that operations while zoomed are appropriately smooth.

floatvsint

This also fixes the scrollbars when zoomed. There's a couple of issues around panning that still need fixing, but they're orthogonal to floating point and needn't wait on it.


Jaex:
You accidentally pushed `diff.txt` file.

EricTetz:
Alright, scrollbar/panning/zoom coordination is fixed.

Previously when you zoomed in/out, the image was panned to the desired center position (client center or mouse), but the CanvasCenterOffset was not updated, so the next time you resized the window or tried to pan using the middle mouse, the image would jump to the wrong offset.

Jaex:
Thanks!

Jaex:
Now I found regression issue with recent changes in panning.

This is before: https://jaex.getsharex.com/2022/02/ShareX_r1Mauw09bj.mp4
This is after: https://jaex.getsharex.com/2022/02/ShareX_T0VJ1J6YyA.mp4

PR #6118, opened 2022-02-20
Fix panning regression (panning at limit accumulates invisible changes)

EricTetz:
This is a fix for [the issue Jaex reported](https://github.com/ShareX/ShareX/pull/6108#issuecomment-1046174097) after the floating point merge.

Minor refactor.  The only routine that updates CanvasCenterOffset now is PanToOffset (replacing AutomaticPan), and this only updates the offset if a pan actually occurred (i.e. the pan wasn't constrained because we're at the pan limit).


Jaex:
I tested it now, it not looks like fixed: https://jaex.getsharex.com/2022/02/ShareX_q9TqC6KQxJ.mp4

Jaex:
You can't reproduce the issue?

Issue #6121, opened 2022-02-22
Zoom: There is no 100% zoom value when started with "Zoom to fit on open" option set


WhiteChairFromIkea:
1) In image editor, set "Zoom to fit on open" option;
2) Zoom using ctrl + mouse wheel. There is no 100% zoom option (see 00:19):

https://user-images.githubusercontent.com/46539653/155109187-800ba5eb-d9e5-4d30-b132-8d03b74b4eb3.mp4

3) Turn off option in step 1, repeat step 2. Now 100% zoom ration is available.

WhiteChairFromIkea:
Yep, this problem exists with latest commits (2022-Feb-23). Ping @EricTetz :)  @EricTetz are you ok to be pinged like that, or I should better leave these issues for sharex team only :) ?

EricTetz:
> @EricTetz are you ok to be pinged like that, or should I better leave these issues (zoom-related) for sharex team only :) ?

Do I get notified of pings in some way? I just happened across this because I saw the issue title.

"Zoom to fit" can result in arbitrary zoom values, which is desirable. The issue then is that the Zoom routine adds or subtracts fixed intervals from there, so if you start off at 113, then zooming in (+25) will give you 138 instead of 125.

I'd recommend having a table of zoom values, then have the zoom routine move to the next highest or lowest in the list. So if you start at 113%, zooming in will go 125, 150, etc., while zooming out will go 100, 90, etc.

WhiteChairFromIkea:
> Do I get notified of pings in some way?

You should get an email, when @EricTetz , y believe.

PR #6127, opened 2022-02-25
Quantize zoom levels.

EricTetz:
Fix for [issue #6121](https://github.com/ShareX/ShareX/issues/6121).

We maintain a list of valid zoom factors. "Zoom to fit" ignores this, and always uses the best value to fit the window, limited only by the min and max zoom factors. Other zoom operations move to the next higher or lower zoom value.

So if we have a zoom of 113%, zoom in will move to 125, 150, 175, etc., zooming out will move to 100, 90 80, etc.


Jaex:
You have GitHub notifications disabled I think, did you check my reply here: #6118

PR #6173, opened 2022-03-23
Update Chinese String of “ScreenRecorder”

wsxjump:
for "ScreenRecorder", it's More accurate to use "屏幕录制" than “屏幕记录”


wsxjump:
And keep consistency with this file:
https://github.com/ShareX/ShareX/blob/a3f615d72caf898b25c7a70cb947ff4196802c1d/ShareX/Forms/MainForm.zh-CN.resx#L327

PR #6176, opened 2022-03-24
Add Vultur (s3 compatible) Block Storage

dutchie027:
Not sure why the diff picked up every line changing, but the main change is the inclusion of S3 type storage from Vultr, similar to Digital Ocean. Their endpoint and AWS S3 Region are included in this


Jaex:
> Not sure why the diff picked up every line changing

I can't accept such PR. Line endings must be different in every lines.

PR #6190, opened 2022-04-04
Switched ctrl+c and ctrl+shift+c behaviors, updated resource files as well

exKitsune:
While using the image editor, I found it much more convenient to have Ctrl + C be the keybind that copies the current state of the image. A friend of mine as well as I sometimes edit a singular photo multiple times (ex. drawing circles around features within an image we want to highlight, but one at a time) and we would like to avoid having to press an extra key to do default copy behavior. 

This should only affect the image editor, and new uses will be able to avoid having info data overwrite their clipboards in region select mode because of muscle memory.


exKitsune:
bump

exKitsune:
bump

PR #6202, opened 2022-04-11
Update Resources.ru.resx

ViktorOn:


L1Q (approved):
Nice catch with those typos :+1:

PR #6223, opened 2022-04-22
fix #6222 ( set IsImageModified to false only on Save and SaveAs OK )

### LINKED ISSUE #6222: Editor window is closing but should not
Reproducing steps:

- Edit an image by opening an editor window 
- Draw a shape on it
- Close the window by clicking X, it shows you a question
- Answer No to the 'not saved changes' dialog message
- Click the Save As button and cancel the dialog box
- Close the window by clicking X, it does NOT shows you the question and close really

laulo13:
Move "IsImageModified = false" 
from 
  ReceiveImageForTask
to 
  OnSaveImageRequested success
and
  OnSaveImageAsRequested success

### DIFF
--- a/ShareX.ScreenCaptureLib/Forms/RegionCaptureForm.cs
+++ b/ShareX.ScreenCaptureLib/Forms/RegionCaptureForm.cs
@@ -1552,8 +1552,6 @@ namespace ShareX.ScreenCaptureLib
         {
             Bitmap bmp = GetResultImage();
 
-            ShapeManager.IsImageModified = false;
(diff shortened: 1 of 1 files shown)


Jaex:
Thanks.

Jaex:
One user reported this now:

> sharex is prompting me about unsaved changes despite copying to clipboard
> since the last update

I was unsure about adding `IsImageModified = false;` to clipboard copy and rest of tasks too. But to keep same functionality as before, decided to add it.

Issue #6228, opened 2022-04-25
Http url for S3 doesn't work


kennedyoliveira:
When using a local s3 server (Minio) and a `http` custom endpoint ShareX will change it to `https` and thus won't work.

For example, I have a Minio server running at the IP and port `192.168.50.123:19000` and I have this configuration for s3:
![image](https://user-images.githubusercontent.com/7723330/165005138-8040a7d4-2b42-4d64-aeee-8a180477d6dd.png)

When running the upload I get this error:

```
Error message:
The underlying connection was closed: An unexpected error occurred on a send.

Request URL:
https://192.168.50.123:19000/sharex/2022/04/nngCYH9UlZ.png

Stack trace:
   at System.Net.HttpWebRequest.GetRequestStream(TransportContext& context)
   at System.Net.HttpWebRequest.GetRequestStream()
   at ShareX.UploadersLib.Uploader.GetResponse(HttpMethod method, String url, Stream data, String contentType, Dictionary`2 args, NameValueCollection headers, CookieCollection cookies, Boolean allowNon2xxResponses)

Upload to Amazon S3 failed.
```

I tried using the `Use custom domain` option, or the endpoint option, and nothing works, both do the same thing, is it possible to make it use the exact URL we configure instead of changing the protocol?

Running version 13.7

PR #6257, opened 2022-05-14
Change url prefix addition in S3 uploader

Jefemy:
Hosting a local version of an S3 compatible service like MinIO doesn't require https, but the uploader forces https no matter what. If the user specifies that it should be uploaded with http then it should use http.

Related issue #6228

### DIFF
--- a/ShareX.HelpersLib/Helpers/URLHelpers.cs
+++ b/ShareX.HelpersLib/Helpers/URLHelpers.cs
@@ -420,6 +420,11 @@ namespace ShareX.HelpersLib
         {
             return URLPrefixes.Any(x => url.StartsWith(x, StringComparison.InvariantCultureIgnoreCase));
         }
+        
+        public static string GetPrefix(string url)
+        {
+            return URLPrefixes.Find(x => url.StartsWith(x, StringComparison.InvariantCultureIgnoreCase));
+        }
 
         public static string FixPrefix(string url, string prefix = "http://")
         {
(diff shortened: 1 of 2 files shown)


Jaex:
I also thought like you and marked that issue as "Bug" and exactly did your change, then right before pushing it realized it not gonna work like that when `isPathStyleRequest` is false. Then I removed "Bug" because it was working as intended and there was reason I did it like that before.

Jefemy:
> I also thought like you and marked that issue as "Bug" and exactly did your change, then right before pushing it realized it not gonna work like that when `isPathStyleRequest` is false. Then I removed "Bug" because it was working as intended and there was a reason I did it like that before.

What about [these changes](https://github.com/Jefemy/ShareX/blob/7c1a0b3ab716ded9c20bf537b56ea6b420bbf84c/ShareX.UploadersLib/FileUploaders/AmazonS3.cs#L128)? I couldn't find a way to get the scheme out of the url using the available URLHelpers so it can't generate a proper `http://` url when using non-path based urls, but it should work with the path style now.

Jaex:
That would be workaround and not real solution.

Jefemy:
> That would be workaround and not real solution.

Added a `GetPrefix` helper to URLHelpers that returns the type of scheme used which is then added back to the host value when url is generated. Should now work in every case.

Jaex:
Thanks.

Jefemy:
Sorry, I used the wrong method. https://github.com/Jefemy/ShareX/blob/29ed48137b8eba84c385a84943ad55d808f492a2/ShareX.HelpersLib/Helpers/URLHelpers.cs#L426 needs to be changed to `URLPrefixes.FirstOrDefault`. Should I make another pull?

Jaex:
I fixed already.

PR #6282, opened 2022-06-04
Romanian translations

Edward205:


Jaex:
Please only pull request new language translation after it is 100% translated. Because this pull request looks like missing a lot of translation files and we can't release new language translation when it is 50% missing.

Edward205:
Could you tell me the files that are missing? I am unfamiliar with this system, so I downloaded many resx files but I may have missed some.

PR #6287, opened 2022-06-08
Chinese translation for last two settings

AnzhiZhang:
This is for these two settings
image
image


mamaruo (commented):
The translation looks good

Jaex:
Entire line endings must be changed if it showing diff in each lines.
This PR also had same problem: #6192
I can't accept PRs when line endings not matching and causing diffs in each lines.

Jaex:
When I select this in GitHub:



It shows correct amount of diffs. But that still proves that line endings must be problematic and your VS code just ignoring them while showing the diff?

AnzhiZhang:
It is correct now, I found something added into ending in git bash:
image
Maybe it's an issue of online vscode, when I open the project in my local vscode, it works well.

PR #6293, opened 2022-06-14
Update Simplified Chinese translation

yfdyh000:
I hope online translation platforms are considered (like Crowdin, Transifex), since touching the .resx source code is not very cool for translators.


Jaex:
Thanks for translation. We tried Crowdin before but resx was very problematic format, that no online translation platform we tried supports it perfectly.

PR #6304, opened 2022-06-23
Romanian translations

Edward205:
For future translators:
Please use the following resources:
https://www.microsoft.com/en-us/language/
https://www.deepl.com/translator

Pentru viitorii traducători:
Vă rog să folosiți următoarele resurse:
https://www.microsoft.com/en-us/language/
https://www.deepl.com/translator


L1Q:
We discourage the use of machine translation. As for tools and resources, please follow our translation guide https://getsharex.com/docs/translation

Edward205:
I find it useful when I know there's a better translation for a word but I cannot remember it, so I use DeepL as kind of a dictionary. I also check the translation before pasting it to make sure it makes sense in the context and there are no mistakes. I consider that the acceptable usage of automatic translation.

L1Q:
Why are you updating non-Romanian files?

Edward205:
I accidentally deleted a key which I have to add back

PR #6320, opened 2022-07-03
Polish translations

RikoDEV:


Jaex:
Thanks for the translation! I was busy with bug fixes after major release therefore couldn't merge it earlier.

PR #6334, opened 2022-07-08
Allowed Zoom feature to work with PavNum keys

Gurgarath:
PavNum keys (0 / + / -) will now work for the zoom feature


Jaex:
Thanks.

PR #6401, opened 2022-07-23
Added an SSL check box for URL

Impdevil:
Added for myself using a nextcloud on a 443 so need to change the initial http to https. only added to nextcloud setup


Jaex:
Why need option for it, you could just add `https://` to URL? But indeed I must change default value of `FixPrefix` function to `https://` instead of `http://`.

Impdevil:
its useful to have the http being the default to be fair, because i would have thought a lot of people have nextcloud sitting on something without a certificate. though if you're just going to change the default to https lol.

Jaex:
`FixPrefix` already does that. So it won't make `http://https://a.url.com`. So I'm not sure how you got that url.

PR #6402, opened 2022-07-24
Change "中止" to "取消"

huangyinhaow:
The meaning of "中止" in Chinese is very similar to stopping (“停止”), but what is expressed here is actually the meaning of "canceling" (“取消”), so it is better to change "中止" to "取消".

### DIFF
--- a/ShareX.ScreenCaptureLib/Forms/ScreenRecordForm.zh-CN.resx
+++ b/ShareX.ScreenCaptureLib/Forms/ScreenRecordForm.zh-CN.resx
@@ -121,7 +121,7 @@
     <value>ShareX - 屏幕录制</value>
   </data>
   <data name="btnAbort.Text" xml:space="preserve">
-    <value>中止</value>
+    <value>取消</value>
   </data>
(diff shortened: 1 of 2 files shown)


huangyinhaow:
图片

I don't know why commits are like these, I just use github.dev to write.

L1Q:
The newline character in your commit is most likely different. You can see the real changes by disabling "whitespace" in diffs.
image
It is still recommended to avoid committing different whitespace because it completely destroys the history of changes for the file.

huangyinhaow:
> The newline character in your commit is most likely different. You can see the real changes by disabling "whitespace" in diffs. image It is still recommended to avoid committing different whitespace because it completely destroys the history of changes for the file.

But why it happened ? Could you tell me?

huangyinhaow:
See https://github.com/ShareX/ShareX/pull/6404

huangyinhaow:
> The newline character in your commit is most likely different. You can see the real changes by disabling "whitespace" in diffs. image It is still recommended to avoid committing different whitespace because it completely destroys the history of changes for the file.

Thanks a lot.

PR #6404, opened 2022-07-24
Change "中止" to "取消"

huangyinhaow:
The meaning of "中止" in Chinese is very similar to stopping (“停止”), but what is expressed here is actually the meaning of "canceling" (“取消”), so it is better to change "中止" to "取消".


Jaex:
Thanks.

PR #6410, opened 2022-07-26
Widen Video Converter Form to fit translated lines

L1Q:
- widen the form by 102 pixels
- widen and move some elements accordingly
- change the default text for lower/higher video quality to represent real world line lengths in the form designer

Before (Romanian):
image
After (Romanian):
image

Before (English):
image
After (English):
image

Designer:
image


Jaex:
Thanks.

Issue #4513, opened 2019-12-06
Feature to snip/crop out/shorten middle of long pictures


nielsmh:
When capturing e.g. long drop-down menus, and most of the menu is irrelevant for the context, it would be convenient with a feature to snip/crop out the _middle_ of the image, with an effect-shape inserted to denote that part of a long/wide image has been removed.

vatterspun:
I'm not sure if @Jaex will do anything about this, but I'm curious and having difficulty picturing what you imagine.  Can you put together a before-and-after of what you have in mind?

nielsmh:
Here's an attempt at doing a quick and dirty job, took perhaps 3-5 minutes in Paint.net, and looks pretty bad.

![image](https://user-images.githubusercontent.com/1062071/71616708-f05f2c00-2bb7-11ea-9b25-4677aa3425b8.png)

vatterspun:
> quick and dirty job, took perhaps 3-5 minutes in Paint.net, and looks pretty bad

I disagree totally -- that's pretty great.  I have up-voted the suggestion.

bartosz-jarmuz:
Yes, that is a greatly missed thing!

chilirock:
+ 1 to add this feature. As far as implementation, it's good to have options. I would prefer a more seemless final result without the jagged/ripped portion visible in the final result, e.g. similar to the post from @yahooshua. Thanks.

chilirock:
It appears that with the number of duplicate issues: #4366, #4513, and #6026 that there is some pent up demand for this functionality. There are probably others out there that would benefit that haven't formally expressed their support.

jwoodard80:
I hate to just be a +1 but, yes.... +1

I used Greenshot for years and changed to ShareX. I've been very happy with it, however, with the amount of captures I take the lack of this feature is extremely glaring. I can try to move windows to a optimum spot for a capture but it's still then an extra 10-15min sometimes trying to edit it and cut out the areas I don't want. 

It is a open source program so I understand that I have no place to gripe. I'm extremely appreciative of the hard work that has gone into this software. It makes me happy to have turned several people on to using it with great results. Personally, however, I'm now having to re-evaluating options (free and paid) to see what is there and if a better workflow is possible. Lack of the ability to "cut out" a section of a capture is the main driver of that. I would MUCH rather donate to this software/group than use something else. 

[...]

PR #6454, opened 2022-08-16
Add Cut Out tool

nielsmh:
Implement #4513 by adding a tool to "cut out" the middle parts of images.

Example of the result:
image
image

Feature checklist:
- [x] Select horizontal or vertical regions in the image to remove
- [x] Confirm or cancel the selection similar to the Crop tool
- [x] The selected region to be removed is highlighted
- [x] Other shapes already on the image are trimmed, moved, or removed, to fit the cut out region
- [x] Support for different effects on the cut edge
- - [x] Zig-zag (sawtooth)
- - [x] Wave 
- - [x] Torn edge
- - [x] ~~Gradient crossfade~~ (dropped)
- [x] UI for selecting and configuring cut effects
- [x] Toolbar button graphics
- [x] Interface strings
- [x] ~~Better handling of transparent or adding background to cut effect areas, as well as option for outline or not on the cut effect~~

Handling of outline or background on the cut out effects might be better left for the user to do with Image Effects or external tools. There are some UI issues with making clear that the cut-out effect area is transparent, but implementing edge effects on this would also be excessively much work, and giving a way to fill the cut out effect similarly has some bad UI affordances.
Giving an option to fill or outline the cut out effect would either use the image effects as-is, which could affect other transparent areas on the image, contrary to what the user might expect, or it would need a large amount of work to make the cut out effect code able to draw/fill that part itself, which seems bad from a cost/benefit view.

I am not very familiar with the code so I would appreciate pointers on any parts that need to be structured differently, commit messages, etc.

### DIFF
--- a/ShareX.HelpersLib/Enums.cs
+++ b/ShareX.HelpersLib/Enums.cs
@@ -213,7 +213,7 @@ namespace ShareX.HelpersLib
         RomanNumeralsLowercase
     }
 
-    public enum CutOutEffectType // Localized (probably)
+    public enum CutOutEffectType // TODO: localize
     {
         None,
         ZigZag,
--- a/ShareX.HelpersLib/Enums.cs
+++ b/ShareX.HelpersLib/Enums.cs
@@ -218,7 +218,6 @@ namespace ShareX.HelpersLib
         None,
         ZigZag,
         TornEdge,
-        Wave,
-        Gradient
+        Wave
     }
 }
--- a/ShareX.HelpersLib/Enums.cs
+++ b/ShareX.HelpersLib/Enums.cs
@@ -212,4 +212,13 @@ namespace ShareX.HelpersLib
         RomanNumeralsUppercase,
         RomanNumeralsLowercase
     }
+
+    public enum CutOutEffectType // Localized (probably)
+    {
+        None,
+        ZigZag,
+        TornEdge,
+        Wave,
+        Gradient
+    }
 }
--- a/ShareX.HelpersLib/Enums.cs
+++ b/ShareX.HelpersLib/Enums.cs
@@ -213,7 +213,7 @@ namespace ShareX.HelpersLib
         RomanNumeralsLowercase
     }
 
-    public enum CutOutEffectType // TODO: localize
+    public enum CutOutEffectType // Localized
     {
         None,
         ZigZag,
--- a/ShareX.HelpersLib/Helpers/ImageHelpers.cs
+++ b/ShareX.HelpersLib/Helpers/ImageHelpers.cs
@@ -231,10 +231,10 @@ namespace ShareX.HelpersLib
                     return bmp;
 
                 case CutOutEffectType.ZigZag:
-                    return bmp;
+                    return TornEdges(bmp, effectSize, effectSize, effectEdge, false, false);
 
                 case CutOutEffectType.TornEdge:
-                    return TornEdges(bmp, effectSize, effectSize * 2, effectEdge, false);
+                    return TornEdges(bmp, effectSize, effectSize * 2, effectEdge, false, true);
 
                 case CutOutEffectType.Wave:
                     return WavyEdges(bmp, effectSize, effectSize * 5, effectEdge);
@@ -1749,7 +1749,7 @@ namespace ShareX.HelpersLib
             return bmp;
         }
 
-        public static Bitmap TornEdges(Bitmap bmp, int tornDepth, int tornRange, AnchorStyles sides, bool curvedEdges)
+        public static Bitmap TornEdges(Bitmap bmp, int tornDepth, int tornRange, AnchorStyles sides, bool curvedEdges, bool random)
         {
             if (tornDepth < 1 || tornRange < 1 || sides == AnchorStyles.None)
             {
@@ -1768,53 +1768,57 @@ namespace ShareX.HelpersLib
 
             if (sides.HasFlag(AnchorStyles.Top) && horizontalTornCount > 1)
             {
-                for (int x = 0; x < horizontalTornCount - 1; x++)
+                for (int x = 0; x < bmp.Width; x += tornRange)
                 {
-                    points.Add(new Point(tornRange * x, RandomFast.Next(0, tornDepth)));
+                    int y = random ? RandomFast.Next(0, tornDepth) : ((x / tornRange) & 1) * tornDepth;
+                    points.Add(new Point(x, y));
                 }
             }
             else
             {
                 points.Add(new Point(0, 0));
-                points.Add(new Point(bmp.Width - 1, 0));
+                points.Add(new Point(bmp.Width, 0));
             }
 
             if (sides.HasFlag(AnchorStyles.Right) && verticalTornCount > 1)
             {
-                for (int y = 0; y < verticalTornCount - 1; y++)
+                for (int y = 0; y < bmp.Height; y += tornRange)
                 {
-                    points.Add(new Point(bmp.Width - 1 - RandomFast.Next(0, tornDepth), tornRange * y));
+                    int x = random ? RandomFast.Next(0, tornDepth) : ((y / tornRange) & 1) * tornDepth;
+                    points.Add(new Point(bmp.Width - tornDepth + x, y));
                 }
             }
             else
             {
-                points.Add(new Point(bmp.Width - 1, 0));
-                points.Add(new Point(bmp.Width - 1, bmp.Height - 1));
+                points.Add(new Point(bmp.Width, 0));
+                points.Add(new Point(bmp.Width, bmp.Height));
             }
 
             if (sides.HasFlag(AnchorStyles.Bottom) && horizontalTornCount > 1)
             {
-                for (int x = 0; x < horizontalTornCount - 1; x++)
+                for (int x = bmp.Width; x >= 0; x = (x / tornRange - 1) * tornRange)
                 {
-                    points.Add(new Point(bmp.Width - 1 - (tornRange * x), bmp.Height - 1 - RandomFast.Next(0, tornDepth)));
+                    int y = random ? RandomFast.Next(0, tornDepth) : ((x / tornRange) & 1) * tornDepth;
+                    points.Add(new Point(x, bmp.Height - tornDepth + y));
                 }
             }
             else
             {
-                points.Add(new Point(bmp.Width - 1, bmp.Height - 1));
-                points.Add(new Point(0, bmp.Height - 1));
+                points.Add(new Point(bmp.Width, bmp.Height));
+                points.Add(new Point(0, bmp.Height));
             }
 
             if (sides.HasFlag(AnchorStyles.Left) && verticalTornCount > 1)
             {
-                for (int y = 0; y < verticalTornCount - 1; y++)
+                for (int y = bmp.Height; y >= 0; y = (y / tornRange - 1) * tornRange)
                 {
-                    points.Add(new Point(RandomFast.Next(0, tornDepth), bmp.Height - 1 - (tornRange * y)));
+                    int x = random ? RandomFast.Next(0, tornDepth) : ((y / tornRange) & 1) * tornDepth;
+                    points.Add(new Point(x, y));
                 }
             }
             else
             {
-                points.Add(new Point(0, bmp.Height - 1));
+                points.Add(new Point(0, bmp.Height));
                 points.Add(new Point(0, 0));
             }
 
--- a/ShareX.HelpersLib/Helpers/ImageHelpers.cs
+++ b/ShareX.HelpersLib/Helpers/ImageHelpers.cs
@@ -223,7 +223,7 @@ namespace ShareX.HelpersLib
             return null;
         }
 
-        public static Bitmap TrimBitmapInteriorHorizontal(Bitmap bmp, int x, int width)
+        public static Bitmap CutOutBitmapMiddleHorizontal(Bitmap bmp, int x, int width)
         {
             if (bmp != null && width > 0)
             {
@@ -255,7 +255,7 @@ namespace ShareX.HelpersLib
             return null;
         }
 
-        public static Bitmap TrimBitmapInteriorVertical(Bitmap bmp, int y, int height)
+        public static Bitmap CutOutBitmapMiddleVertical(Bitmap bmp, int y, int height)
         {
             if (bmp != null && height > 0)
             {
--- a/ShareX.HelpersLib/Helpers/ImageHelpers.cs
+++ b/ShareX.HelpersLib/Helpers/ImageHelpers.cs
@@ -1683,6 +1683,7 @@ namespace ShareX.HelpersLib
                 using (TextureBrush brush = new TextureBrush(bmpIn))
                 {
                     g.SetHighQuality();
+                    g.PixelOffsetMode = PixelOffsetMode.Half;
                     g.FillPolygon(brush, path);
                 }
                 return bmpResult;
@@ -1829,6 +1830,7 @@ namespace ShareX.HelpersLib
             using (TextureBrush brush = new TextureBrush(bmp))
             {
                 g.SetHighQuality();
+                g.PixelOffsetMode = PixelOffsetMode.Half;
 
                 Point[] fillPoints = points.Distinct().ToArray();
 
--- a/ShareX.HelpersLib/Helpers/ImageHelpers.cs
+++ b/ShareX.HelpersLib/Helpers/ImageHelpers.cs
@@ -1673,6 +1673,8 @@ namespace ShareX.HelpersLib
             int horizontalWaveCount = Math.Max(2, (bmp.Width / waveRange + 1) / 2 * 2) - 1;
             int verticalWaveCount = Math.Max(2, (bmp.Height / waveRange + 1) / 2 * 2) - 1;
 
+            int step = Math.Min(Math.Max(1, waveRange / waveDepth), 10);
+
             Bitmap updateResult(Bitmap bmpIn, Point[] path)
             {
                 Bitmap bmpResult = bmpIn.CreateEmptyBitmap();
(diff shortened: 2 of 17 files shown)


ShareX.HelpersLib/Enums.cs
```diff
@@ -212,4 +212,13 @@ public enum StepType // Localized
         RomanNumeralsUppercase,
         RomanNumeralsLowercase
     }
+
+    public enum CutOutEffectType // Localized (probably)
```
Jaex, line 216: It is not localized yet.

ShareX.HelpersLib/Helpers/ImageHelpers.cs
```diff
@@ -223,6 +223,126 @@ public static Bitmap CropBitmap(Bitmap bmp, Rectangle rect)
             return null;
         }
 
+        public static Bitmap CutOutBitmapMiddleHorizontal(Bitmap bmp, int x, int width, CutOutEffectType effectType, int effectSize)
```
Jaex, line 226: You can add `Orientation orientation` parameter like in `CombineImages` function instead of using two different functions for horizontal and vertical. Because majority of codes are shared between these two functions.

ShareX.ScreenCaptureLib/Shapes/Tool/CutOutTool.cs
```diff
@@ -0,0 +1,188 @@
...
+        public bool IsHorizontalTrim => Rectangle.Width >= Options.MinimumSize && Rectangle.Width > Rectangle.Height;
+        public bool IsVerticalTrim => Rectangle.Height >= Options.MinimumSize && Rectangle.Height >= Rectangle.Width;
+
+        public override bool IsValidShape => IsHorizontalTrim || IsVerticalTrim;
+
+        private ImageEditorButton confirmButton, cancelButton;
+        private Size buttonSize = new Size(80, 40);
+        private int buttonOffset = 15;
+
+        private Brush EffectBrush = new SolidBrush(Color.FromArgb(128, Color.Gray));
```
Jaex, line 47: Brush must be disposed.

ShareX.ScreenCaptureLib/Shapes/Tool/CutOutTool.cs
```diff
@@ -0,0 +1,188 @@
...
+                    cancelButton.Rectangle = confirmButton.Rectangle.LocationOffset(0, buttonSize.Height + buttonOffset);
+                }
+            }
+        }
+
+        public override void OnDraw(Graphics g)
+        {
+            if (IsHorizontalTrim)
+            {
+                g.FillRectangle(EffectBrush, new RectangleF(Rectangle.X, g.ClipBounds.Y, Rectangle.Width, g.ClipBounds.Height));
```
Jaex, line 112: I wonder how it would look if we used transparent checker pattern using texture brush, like this:



Because currently in gray background it can be hard to see selection, because selection is also gray.

nielsmh: I'm not entirely sold on it, but on the other hand maybe it does better denote "this part is getting tossed out".

image

Jaex: I mean't like this:

```
using (Image selectionHighlightPattern = ImageHelpers.CreateCheckerPattern(1, 1, Color.FromArgb(128, Color.LightGray), Color.FromArgb(128, Color.Gray)))
```

ShareX.HelpersLib/Helpers/ImageHelpers.cs
```diff
@@ -1596,6 +1664,91 @@ public static void FastBoxBlur(Bitmap bmp, int radius)
             }
         }
 
+        public static Bitmap WavyEdges(Bitmap bmp, int waveDepth, int waveRange, AnchorStyles sides)
```
Jaex, line 1667: You should add this as image effect too, it is cool!

Jaex: By the way wouldn't this option also work?

ShareX.ScreenCaptureLib/Shapes/Tool/CutOutTool.cs
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        public override void OnDraw(Graphics g)
+        {
+            using (Image selectionHighlightPattern = ImageHelpers.CreateCheckerPattern(8, 8, Color.FromArgb(128, Color.White), Color.FromArgb(128, Color.Gray)))
+            using (Brush selectionHighlightBrush = new TextureBrush(selectionHighlightPattern, System.Drawing.Drawing2D.WrapMode.Tile))
+            {
+                if (IsHorizontalTrim)
+                {
+                    g.FillRectangle(selectionHighlightBrush, new RectangleF(Rectangle.X, Math.Max(g.ClipBounds.Y, Manager.Form.CanvasRectangle.Y), Rectangle.Width, Math.Min(g.ClipBounds.Height, Manager.Form.CanvasRectangle.Height)));
```
Jaex, line 126: These rectangles can be stored as property such as:

```cs
        public Rectangle CutOutRectangle
        {
            get
            {
                ...
            }
        }
```

That way you can also use it inside `OnUpdate()` method while positioning buttons properly.

ShareX.HelpersLib/Helpers/ImageHelpers.cs
```diff
@@ -1768,53 +1768,57 @@ public static Bitmap TornEdges(Bitmap bmp, int tornDepth, int tornRange, AnchorS
...
-                    points.Add(new Point(tornRange * x, RandomFast.Next(0, tornDepth)));
+                    int y = random ? RandomFast.Next(0, tornDepth) : ((x / tornRange) & 1) * tornDepth;
+                    points.Add(new Point(x, y));
                 }
             }
             else
             {
                 points.Add(new Point(0, 0));
-                points.Add(new Point(bmp.Width - 1, 0));
+                points.Add(new Point(bmp.Width, 0));
```
Jaex, line 1777: Wouldn't `bmp.Width` be outside of bitmap? Because index is 0 based.

ShareX.ScreenCaptureLib/Shapes/Tool/CutOutTool.cs
nielsmh: That does make the code a bit neater, but I don't see how it'd help in OnUpdate(). The button positioning barely concerns itself with the specific shape of the entire cut out area, just the two extents in the direction being cut, and those are the same for the user selection and the displayed cut. I think using the user selection (`Rectangle` property) is better for the buttons, since that will typically position them closer to where the user releases the mouse.

The OnUpdate() button positioning could be shortened a bit, but that would make it harder to read/follow, so I think I prefer keeping it as-is.

ShareX.ScreenCaptureLib/Shapes/Tool/CutOutTool.cs
```diff
@@ -0,0 +1,203 @@
...
+    public class CutOutTool : BaseTool
+    {
+        public override ShapeType ShapeType { get; } = ShapeType.ToolCutOut;
+
+        public override bool LimitRectangleToInsideCanvas { get; } = true;
+
+        public bool IsHorizontalTrim => Rectangle.Width >= Options.MinimumSize && Rectangle.Width > Rectangle.Height;
+        public bool IsVerticalTrim => Rectangle.Height >= Options.MinimumSize && Rectangle.Height >= Rectangle.Width;
+
+        public override bool IsValidShape {
```
Jaex, line 42: Tip:

ShareX.HelpersLib/Helpers/ImageHelpers.cs
nielsmh: It doesn't produce errors. Also I'm pretty sure the paths are left-inclusive right-exclusive (and similar top-inclusive bottom-exclusive).

Jaex:
> Better handling of transparent or adding background to cut effect areas, as well as option for outline or not on the cut effect

What about:

```cs
ImageHelpers.FillBackground(img, Options.EditorCanvasColor);
```

Which is transparent by default.

ShareX.HelpersLib/Helpers/ImageHelpers.cs
nielsmh: image

I do wonder if it would make sense to integrate some drop shadow or outline-drawing in the cut out effect, or at least a way to draw it on a non-transparent background. Right now it "lies" in the editor where it displays as blackness in the cut-out effect where it's actually transparent.

ShareX.HelpersLib/Helpers/ImageHelpers.cs
Jaex: You can draw outside, it won't throw error.

> left-inclusive right-exclusive

What is that even means.

ShareX.HelpersLib/Helpers/ImageHelpers.cs
Jaex: I don't think we should add drop shadow.

Jaex: What you mean it lies and displays blackness? I literally draw black there in my screenshot myself to make it easily visible.

ShareX.HelpersLib/Helpers/ImageHelpers.cs
nielsmh: I tried with and without -1 on the far edges. The result is definitely more wrong with it than without it, but it's wrong even without it too.
For the first two of these images, I cut out the top edge of the image with Wave Edge effect. The third image was opened in the image editor and then saved with no changes made. The first has -1 on Width and Height coordinates for the effect-less edges, the second uses Width and Height as given.
After saving the images from ShareX, I opened them in Krita, added a fill layer underneath, and saved again.

It seems that the pixel coordinate model is in fact center of pixel, so (0, 0) is the center of the top left pixel, meaning that a rectangle starting at (0, 0) excludes half of the top row and half of the left row of pixels. For correct, seamless handling (with float-coordinates only?) the top left of the top left pixel then has to be addressed as (-0.5, -0.5), and the right edge of the rightmost pixel has coordinate (Width - 0.5).

devenv_2022-08-17_23-52-46 bg
devenv_2022-08-17_23-53-42 bg
devenv_2022-08-17_23-57-40 bg

ShareX.HelpersLib/Helpers/ImageHelpers.cs
nielsmh: I did try using the CurvedEdges flag for TornEdges before writing the WavyEdges function. It doesn't give the same kind of uniform result that WavyEdges does, where the two sides of the cut perfectly follow each other, so both have their uses depending on the "mood" the user wants to convey.

ShareX.HelpersLib/Helpers/ImageHelpers.cs
nielsmh: Yep, offsetting everything by -0.5 makes it hit the image edges perfectly.

image

ShareX.HelpersLib/Helpers/ImageHelpers.cs
nielsmh: > What you mean it lies and displays blackness? I literally draw black there in my screenshot myself to make it easily visible.

In the image editor it looks like this:
image

Like the cut-out effect is a black river.
But when you save the image as PNG, or add effects to it that use transparency, it becomes clear that the "black river" is in fact a transparent area.

nielsmh: Okay, when using a large Torn Edge effect that leaves "big lakes" instead of just a "narrow river", it becomes slightly visible that the black area is in fact a very dark checkerboard indicating transparency. That's very difficult to tell when it contrasts hard with the surrounding UI and/or the user's image being edited. I would suggest that the checkerboard effect used depends on the user's UI preference for light/dark.

nielsmh:
I've decided that adding any kind of integrated outline or background-fill for the cut out effects, directly in the tool, is probably not worth it.
There are two approaches to doing it, and both have some bad disadvantages.
1. Naively using the outline and background fill effects from the helper library, after performing the cut. Give the user UI options for selecting outline size and color, and background fill, in the cut out tool settings. This has the issue that it will end up affecting _all_ transparent areas in the image, not just the part that was just made by this tool. That would likely be contrary to user expectations, and could give some unfortunate cases of destroyed work. The user can just as well do this on their own using the Image Effects function from within the image editor.
2. Implementing outlining and background filling as part of the WavyEdges, TornEdges effects. This would suddenly make the code a lot more complex, and would also need to be very careful about handling pre-existing transparent areas in the image. It would be a lot of implementation work that would make the code much more complex, having to handle lots of corner cases (and some of those might have many "correct" solutions, depending on situation and user goals), and the real benefit might not stand measure to the effort required.

So the first is scrapped because it would lead to UI confusion and betrayal of user expectations, and the second is scrapped because it looks bad from a cost/benefit point of view, and might still hit some of the issues the first also has.

ShareX.HelpersLib/Helpers/ImageHelpers.cs
Jaex: https://github.com/ShareX/ShareX/search?q=PixelOffsetMode

ShareX.HelpersLib/Helpers/ImageHelpers.cs
Jaex: > I would suggest that the checkerboard effect used depends on the user's UI preference for light/dark.

It already does:

nielsmh: I have no idea what was wrong with my settings then, but toggling custom theme on and back off fixed it.

ShareX.HelpersLib/Helpers/ImageHelpers.cs
nielsmh: That certainly makes things easier.

The coordinates still need to use bmp.Width and bmp.Height to be correct and not leave a one-pixel edge on the right and bottom.

(On a side note, I find it weird that Microsoft never directly documents what the pixel coordinate system for antialiasing is, or at least I can't find it mentioned. Oh well.)

chilirock:
Hi @nielsmh. Thank you so much for taking the lead on building this new capability. I can't wait for it to be included in a future release. 

Questions: 

Will this new function allow multiple successive "cut outs" or do all the cut outs have to be completed in one selection in one image edit session? For example, the user might like to cut out a horizontal strip, see the resulting modified image before cutting out a second horizontal strip, or a vertical strip. 

Can you provide a little more description of the region selection handles will work. In your screenshot above, it appears that the handles would allow users to select a region that is not the full width or height of the image. Is that the case, or do they just allow selection of full width or height? Are the width and height selected and cut independently?

Where in the Image Editor menu will the "cut out" function be called from?

Jaex:
You can get answer to your all questions by using [dev build](https://getsharex.com/docs/dev-builds).

Jaex:
Maybe adding "Quick crop" option can make sense? Which would crop instantly without show confirmation buttons.

nielsmh:
> I would suggest you to not use "force push" though. Because every time someone does force push, then I must re check everything to make sure nothing else is modified, instead of just recent commit.

Another project I participate in puts a lot of emphasis on commit history and making every commit look good on its own, and prefers force push to clean up messy commit histories and avoid commits that get partly or wholly undone by a later one in the same PR. So it's a bit of a habit, sorry.


> Hi @nielsmh. Thank you so much for taking the lead on building this new capability. I can't wait for it to be included in a future release.

I hope you'll like how it turned out. I'm looking forward to be able to use it in my own work just as much :) 

> Will this new function allow multiple successive "cut outs" or do all the cut outs have to be completed in one selection in one image edit session? For example, the user might like to cut out a horizontal strip, see the resulting modified image before cutting out a second horizontal strip, or a vertical strip.

You can do multiple cuts after each other on the same image, or you can load an existing image into the image editor and do cuts on that too. There isn't any limit on that.


> Can you provide a little more description of the region selection handles will work. In your screenshot above, it appears that the handles would allow users to select a region that is not the full width or height of the image. Is that the case, or do they just allow selection of full width or height? Are the width and height selected and cut independently?

The selection handles are removed in the final version of this PR. The only way to resize the selection is to draw a new from scratch. You _can_ grab the selection rectangle (if you can find it) and move it around, and it will resize if you try to drag it over the image edges, but that's just a weird side effect. Maybe that should also be disabled somehow.
It's only possible to cut out in the full width or height.

> Where in the Image Editor menu will the "cut out" function be called from?

It's a tool next to the Crop button.


> Maybe adding "Quick crop" option can make sense? Which would crop instantly without show confirmation buttons.

Something like holding Shift or Ctrl while dragging out the selection, maybe.

chilirock:
> You can do multiple cuts after each other on the same image

That is Great, thanks. 

> The selection handles are removed in the final version of this PR. The only way to resize the selection is to draw a new from scratch

Since there won't be any selection handles, it could be hard to get selection just right with just the mouse drag and release. It would be good if there was a way to fine tune the start/end of the selection either by using the arrow keys or by moving the start/end of the selection area before confirming the cut.

> It's a tool next to the Crop button.

Makes sense to put it in the main toolbar next to Crop. Will you also add it to the "Image" drop down menu next to Crop? see below

image

Jaex:
> Something like holding Shift or Ctrl while dragging out the selection, maybe.

I also thought about it and decided it is not better than option. I never need confirmation, so option would be better. Which won't be enabled by default of course.

Jaex:
I noticed regression issue in "Torn edge" effect, which probably also affects "Wave edge" effect.

Before: https://jaex.getsharex.com/2022/08/ShareX_sb6PxKZ7cK.mp4
After: https://jaex.getsharex.com/2022/08/ShareX_ARyeEwXkHW.mp4

As you can see from recordings, edges are not connected properly anymore.

By the way you may also want to add "Zig zag edge" effect.

nielsmh:
Honestly I'd prefer to say "Don't do that" to using settings that cause weird corner effects, but I do have some ideas on how it can be fixed. I'll make a new PR when I have a fix ready for testing.

PR #6548, opened 2022-10-11
Use editor DPI when rendering final image

### LINKED ISSUE #6427: Improper scaling of fonts on save in 14.1.0
When saving an image edited with ShareX, first resized down and annotated fonts on saved fine are much bigger, than displayed in the editor.

Steps to reproduce:

1. Open hi resolution image. I used 4000x3000 px.
2. Resize it down to 960x720 px.
3. Annotate with Arial, 24.
4. Save the image file.
5. Compared what you edited and what has been stored.
6. Saved image should match the one displayed before saving, but it doesn't. Fonts are much bigger and don't fit into their boxes.

L1Q:
Fixes #6427


Jaex:
Thanks.

PR #6603, opened 2022-11-23
Color tools enhancements

### LINKED ISSUE #6602: Color tools usability issues
The color tools currently lacks some minor usability improvements. Current issues include:

- Incorrect tab indexes on the ColorPicker form controls, making it difficult to change correlated control values with keyboard only. Hitting tab while "R" is focussed should set focus to "G", etc.
- Text input control values should have their text selected when the control receives focus, to accommodate typing in a new value or changing the current one.
- Hex value control ignores alpha value (this is ok if there is no opacity, but if there is, hex should display the RGBA value not RGB value)
- Defaulting to 0 first when toggling the alpha control seems unintuitive; if partially opaque, clicking it should reset to 255 first, then toggle to 0 on the second click - the inverse behavior it currently has.
- No support for short form color values (`#F00` is a valid color for red, and should be recognized as `#FF0000` if input into the hex field).

I have fixed the above issues on my [color-tools-enhancements branch](https://github.com/StelioKontosXBL/ShareX/tree/color-tools-enhancements) and I can open a PR with the changes for review.

Ste1io:
Fixes #6602 with numerous usability and functionality improvements.

Adds support for short form hex colors, fixes tabindex for related color groups and their input controls, pre-selects the value of a textinput field when it receives focus, and adds support for RGBA in hex.


StelioKontosXBL:
Bump to escape the stale reaper. @Jaex

Ste1io:
Can we get this reopened, @Jaex ?

PR #6796, opened 2023-03-07
Add Cache Control settings to Azure Uploader

Scrxtchy:
The documentation covering this isn't straightforward on azure docs, but I gave it a shot anyway.
We can't apply cache-control to an entire storage container, so we have to do these per file, therefore putting it as a part of the upload process is necessary unless you want to do this through a separate rest request/azure cli

Adds a text box to the uploader config for users to specify their header
image

These are implemented directly onto the blob
23-03-08_03-56-00-20230323-03-08_03-39-08-tiendasubirAzure php_at_3e


Jaex:
Thanks!

PR #6872, opened 2023-05-01
Move recorder user arguments to avoid overwrites

Kurochi51:
Previously user arguments were appended near the beginning of the line, which resulted in any pre-existing arguments to overwrite them. Now the arguments are appended just before the output, allowing users to actually overwrite any hardcoded options.


Kurochi51:
Added additional checks in the video encoder portion to better handle identical arguments, and remove hardcoded ones if the user decides to pass the same argument. This way it removes some of the ambiguity on which argument ffmpeg will respect.

PR #6901, opened 2023-05-27
Better screen capture with ffmpeg ddagrab.

PJB3005:
ddagrab is a relatively new filter video source in ffmpeg that uses DXGI Desktop Duplication (Windows 8+) to do screen recording.

This has some advantages over the existing screen-capture-recorder method:

* Doesn't need installation of a separate driver.
* Doesn't induce stutter in whatever you're recording.

It does mean you'll need a relatively new (within the last year) version of ffmpeg.

DXGI Desktop Duplication works per monitor ("output") which means that I need to stitch together multiple monitors if the capture area is across them. I do this with a filter graph ffmpeg-side. Sadly, they aren't synchronized nicely, so you can see the tearing if you move something across a monitor boundary. 

Two TODOs with the current implementation:

* I have not the slightest of a clue how custom ffmpeg args play into this.
* I'm not sure the code to get DXGI output numbers is correct. Ideally I'd check against the DXGI API directly, but that would mean pulling in some sort of dependency to call DXGI from C#. Let me know if that's ok.

### DIFF
--- a/ShareX.MediaLib/FFmpegCLIManager.cs
+++ b/ShareX.MediaLib/FFmpegCLIManager.cs
@@ -43,6 +43,11 @@ namespace ShareX.MediaLib
         public const string SourceVideoDevice = "screen-capture-recorder";
         public const string SourceAudioDevice = "virtual-audio-capturer";
 
+        /// <summary>
+        /// Uses the <c>ddagrab</c> ffmpeg video source filter.
+        /// </summary>
+        public const string SourceDxgi = "DXGI";
+
         public const int x264_min = 0;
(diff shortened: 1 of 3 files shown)


Jaex:
There is a reason I did not add ddagrab support myself:



Because I did not compile FFmpeg 6.0 and add it to ShareX yet:

https://github.com/ShareX/FFmpeg/releases

PJB3005:
I made sure the code only lists the ddagrab option if the selected version of ffmpeg supports it (for people using custom ffmpeg). Would it make sense to support it either way, before you ship a new version of ffmpeg?

PR #6973, opened 2023-07-03
Implemented some best practices in a few classes

Lehonti:
For example:

- Added null checks (to ascertain that a class has been built correctly)
- Made SnapSize class immutable (its properties are never modified, so removing its mutability makes the code safer and easy to understand)
- Removed unused setters (some properties are never modified. Removing setters makes it easier to track changes, as the developer doesn't have to wonder if a certain property is modified somewhere)
- Removed nesting levels

This is just a small part of what could become an ongoing effort to improve ShareX's codebase


Jay-Tarzan-jojo:
Open

Lehonti:
Can the failing jobs be run again, please? It appears that there's always a timeout somewhere (every time in a different job)

PR #6986, opened 2023-07-11
Refactoring ShareX.IndexerLib

ahmed0saber:


ShareX.IndexerLib/Forms/DirectoryIndexerForm.cs
```diff
@@ -98,18 +98,12 @@ private async Task IndexFolder()
...
-                            wbPreview.Visible = false;
-                            txtPreview.Visible = true;
-                            txtPreview.Text = Source;
-                        }
+                        bool isReadyForWebPreview = Settings.Output == IndexerOutput.Html;
+
+                        txtPreview.Visible = !isReadyForWebPreview;
+                        wbPreview.Visible = isReadyForWebPreview;
+
+                        (isReadyForWebPreview ? wbPreview.DocumentText : txtPreview.Text) = Source;
```
ficcialfaint, line 106: doesn’t seem to be a valid syntax

ahmed0saber: Hello @ficcialfaint , you're right, thanks for letting me know, that was my mistake to use JavaScript syntax in a C# code, so I have reverted changes in this commit.

ficcialfaint: Got it, just saw that the tests are failing and then saw this

PR #7070, opened 2023-09-05
Added support individual hex color channels for color picker parser

yevhenii-sir:
The primary need was to have the ability to obtain BGR hex colors.


Jaex:
What about `rhex` and `rHEX`? So it can be compatible with `r255` and `r1` options.

PR #7108, opened 2023-10-14
OneDrive Add Use Direct Link

Yi-pixel:
I have added the ability to generate direct links for OneDrive (using OneDrive's "Embed" feature). I noticed that there is an implementation for embedding in the source code for OneDrive, but it seems that this is not being called. Therefore, this pull request introduces a new switch to control it.

Additionally, after adding the switch control and relocating it, I observed that there have been numerous changes in ShareX.UploadersLib/Forms/UploadersConfigForm.resx. I'm not sure if this is normal. :D

Its translation is entirely based on other similar switch options.


Jaex:
Thanks.

Issue #7109, opened 2023-10-15
Can we use current image as background image


alimertcetin:
For example, I captured a region and I want an image effect as follows; the captured region on the background (blurred) and the also on the foreground (not blurred) but smaller than background. Is it possible to achieve? Because when I try to add background image on effects section, it only accepts a predefined path. Adding something like "**use source**" toggle would be very helpful.

I captured this:
![0GwaldCrrC](https://github.com/ShareX/ShareX/assets/75283633/cf8bc14e-4a9e-447f-9d19-780b5524fe55)

I want it to be look like in the preview but instead of adding blur for original picture I want to add blur for background:
![ShareX_EbP4mLQRvw](https://github.com/ShareX/ShareX/assets/75283633/a5b932d6-3cae-4313-bb96-a3d875fb6b45)

PR #7110, opened 2023-10-15
Add DrawSourceImage as ImageEffect

alimertcetin:
This allows the user to use the current image without providing a location.
Fixes : #7109 

ShareX_RjPTv8WPJ5


WildFyr16:
How did you get the "SourceImage" node? Or is that something you programmed for this demonstration?

alimertcetin:
> How did you get the "SourceImage" node? Or is that something you programmed for this demonstration?

Yes. I really liked the tool but It was missing a future. When you took a screen shot, there is no path to that picture because it is not saved to disk, you have to save it first in order use it between layers(effects). I changed that by making changes on the core of the program.

Issue #6725, opened 2023-01-27
(UI/Minor) Unsaved changes dialogue has "Yes" and "No" buttons in the wrong place


GregariousJB:
Nitpicky? Actually, technically no. [Jakob's Law](https://lawsofux.com/jakobs-law/) in UX design is pretty clear about doing things that users expect, and this has made me click the wrong box more than once:

![sigh](https://i.imgur.com/KAaihMO.jpeg)

I recommend including a third "Cancel" button as well, just like every other program. It would also be slightly better to change the "Yes" to a "Save" so we can tell when the button does at a quick glance without needing to read about it above. [Why read words when you can click the stuff the words tell you stuff about?](https://www.youtube.com/watch?v=x3V11tRsR3w)

Do what people expect - what they're familiar with. That's "UX 101" in all training guides. Please and thank you 👍

PR #7188, opened 2023-12-18
Implementing save changes before closing dialog for the image editor as requested on issue 6725

gregorygregio:
Altered the message box that is presented when closing the image editor without saving changes to ask if the user wants to save them or not, also giving the cancel option to prevent closing the editor. Might need some work in the translations.
Captura_alteracao_ShareX

fixes: [#6725](https://github.com/ShareX/ShareX/issues/6725)


Jaex:
How did you update all translations?

gregorygregio:
I knew how to translate it to english, spanish and portuguese - which is my native language. All other languages were translated using Google translator.

Jaex:
Sorry but I don't trust Google translator, which I also mentioned at here long time ago: https://getsharex.com/docs/translation

gregorygregio:
I understand. So should I remove the translations? Because the current message would be misleading on what the "Yes" button does.

Jaex:
If text should be changed then you should create a new string resource instead of using old resource.

gregorygregio:
Got it. I will start a new PR from scratch and abandon this one. 
The new PR is #7204

Issue #4079, opened 2019-04-15
[Feature Request] Undo/Redo function on image editor


jacquen:
Hi

Can we add undo/redo function on image editor to support all operation in image editor, specially will be useful by corp and image resize operation, curently there is no way reverse back.

Thanks

gpion:
Duplicate of #3673 ?

jacquen:
not exactly same, if we can have redo/undo function in image resize or corp, even it will be better if we can have this function in image editor globally,  it will help lot.

Caiuzx:
there is still no redo function and all times  I lose my precious time: P

derekantrican:
+1 - could really use a "Ctrl+Y" redo function

rogeriodec:
Undo is working with Ctrl+Z. 
But I don't know why there is no redo (Ctrl+Shift+Z)...

Slapbox:
Undo without redo is just adding another delete button to the application but not labeling it as such :(

Otherwise, incredible app - wow.

3rady:
@rogeriodec wrote:

> Undo is working with Ctrl+Z.

The Undo (Ctrl-Z) function works with many operations, but not all.  For example, the OP mentions the crop operation cannot be undone.

NAL100:
Definitely need undo and redo. Came here to make this a new issue but I spent 20 minutes highlighting and image only to find out I couldn't undo edits. Huge problem that almost makes it unusable.

PontiacCZ:
Oh, such a trivial issue opened from 2019? :-o I came here because I did a crop and was not able to undo it using CTRL + Z . 

Isn't this project abandoned? I have just noticed those 450 opened issues...

Vampire:
Shift + Del can also not be undone :-(

Slapbox:
The fact this issue still exists is so sad, because this is otherwise the greatest friggin' software...

frizzr:
We all love you ShareX developers, but please oh please give us some kind of redo capability here to undo the undo we did in error that leads us to repeat a _lot_ of lost editing work.

I confess I do find the undo operation very, _ahem_, **underwhelming** when it comes to granularity of actions (I'm screaming on the inside here...but with total love for the developers!) 

Seriously though, giving us just a very simple redo action here would save me _literally an hour a month_...and I'm just one person (who still loves your product dearly!)

[...]

PR #7193, opened 2023-12-23
Implemented Undo/redo function on image editory

gregorygregio:
I have reworked the existing implementation of the undo (CTRL+Z) funtion and implemented a redo (CTRL+Y) one.
First I created the ImageEditorMemento class to store a state of the Image being edited, by keeping  the list of shapes drawn and the Bitmap canvas. I decided to have a constructor with only Shapes in order to avoid cloning and storing the whole Image all the time.
image

Next I added two stacks of this ImageEditorMemento class in to the shape manager to keep all the past states of the editor and the undone ones.
image

Then I have created two methods that create and stack the current state of the editor. One will store just the Shapes list and the other will store both the Shapes and the canvas.
image

Finally I have added the Undo and Redo methods, that will unstack the past state or the undone state and restore them in the Shapes list and updating the canvas when it is not null.
image

The previous UndoShape method was replaced by the Undo. 

Also a menu item was added for the Redo function under the Undo one.

fixes: #4079

### DIFF
--- a/ShareX.ScreenCaptureLib/ImageEditorHistory.cs
+++ b/ShareX.ScreenCaptureLib/ImageEditorHistory.cs
@@ -1,0 +1,131 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ShareX.ScreenCaptureLib
+{
+    internal class ImageEditorHistory : IDisposable
+    {
+        private Stack<ImageEditorMemento> mementoStack = new Stack<ImageEditorMemento>();
+        private Stack<ImageEditorMemento> redoMementoStack = new Stack<ImageEditorMemento>();
+        private readonly ShapeManager shapeManager;
+        public ImageEditorHistory(ShapeManager shapeManager)
+        {
+
+            this.shapeManager = shapeManager;
+        }
+
+        public void AddMemento(ImageEditorMemento memento)
+        {
+            mementoStack.Push(memento);
+            foreach (var mmnt in redoMementoStack)
+            {
+                mmnt.Dispose();
+            }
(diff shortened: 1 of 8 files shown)


Jaex:
ImageEditorMemento must have [IDisposeable](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable) and memento list must be disposed when Image Editor is closing. Also while consuming the stacks make sure those mementos are disposed too.

For example you are doing `redoMementoStack.Clear();` but that list must be disposed before clearing to make sure all bitmaps are disposed. Otherwise it will be up to garbage collector to eventually dispose it. But it is better to dispose Bitmap yourself than letting garbage collector do it.

Jaex:
Also you may want to have "ImageEditorHistory" class and store two list in it. Then you can implement [IDisposeable](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable) to that class too. And then add Undo & Redo functions in it. That way it would be cleaner code.

gregorygregio:
Commited the requested changes. The memento class now implements IDisposable and is disposed when the editor is closed.
I have created the ImageEditorHistory class to encapsulate the stacks and the undo/redo behavior.

However I can't dispose the memento that was poped from the stack beacuse it's shapes list is being assigned to the ShapeManager.Shapes list, so if I dispose it, all shapes in the editor would be deleted. Unless I duplicate the shapes list again before assigning it to the ShapeManager.Shapes list, but I am not sure if that would be necessary, since the shapes would be disposed onde they are not used by the editor anymore. What do you think about it?

Jaex:
Great work!

Jaex:
Past 4 hours I'm fixing undo/redo related bugs but more and more appearing. And now I realized that when you pan/zoom/change editor window size etc. then undo inserts shapes in wrong positions and fixing this gonna be pretty complicated :(

PR #7207, opened 2024-01-05
fix: change from http to https on all api endpoints

daniiiol:
All API endpoints have transitioned from HTTP to HTTPS. This change is implemented to ensure the encryption of data in transit, thereby improving the overall security posture of ShareX.

Many services may redirect to https, but this is only an assumption. This change generally relies on secure data transmission.


daniiiol:
I could do another rebase. But then I would need to know whether the PR is also being checked. What were the reasons for not taking the PR into account?

Issue #6653, opened 2022-12-23
Windows on ARM support


Wunkolo:
The previous issue #4922 is pretty old and captures a different time in the Windows-On-Arm ecosystem and I wanted to re-capture interest in having _native_ support for ARM64-Windows.

I've had some luck updating the project to net framework 4.8.1 and adding a custom "ARM64" target and trying to build with the command `msbuild -m -p:Configuration=Debug -p:Platform="ARM64" -p:Runtimeidentifier="win-arm64"`.
Currently the projects use 4.8.0 and 4.8.1 in particular adds [native support for ARM64](https://github.com/microsoft/dotnet/blob/main/releases/net481/README.md).
Though there are some other issues such as Steam integration supporting x86-targets only:

```
warning MSB3270: There was a mismatch between the processor architecture of the project being
       built "" and the processor architecture of the reference "Steamworks.NET, Version=7.0.0.0, Culture=neutral, proces
       sorArchitecture=x86", "x86". This mismatch may cause runtime failures. Please consider changing the targeted proce
       ssor architecture of your project through the Configuration Manager so as to align the processor architectures bet
       ween your project and references, or take a dependency on references with a processor architecture that matches th
       e targeted processor architecture of your project.
```

and some other project configuration issues due to not anticipating alternate platforms:
```
Your project file doesn't list 'win-arm64' as a "RuntimeIdentifier". You should add 'win-arm64' to t
       he "RuntimeIdentifiers" property in your project file and then re-run NuGet restore.
```

At some point, it will have to download an ARM64-build of ffmpeg in which an implementation can be referenced from here:
https://github.com/dvhh/ffmpeg-wos-arm64-build

huangyinhaow:
+1

Wunkolo:
[A PR was made](https://github.com/ShareX/ShareX/pull/7228) that implements this but seems to have been automatically closed by the GitHub actions bot. It's certainly worth revisiting as Windows on Arm has been getting some big pushes lately.

InsanityOnABun:
Chiming in as well, ARM builds would be greatly appreciated. x64 builds seem to work through Prism but are very slow.

worldsdream:
+1

worldsdream:
Does anyone know if sharex works on a arm64 device? I know it’s not native. But will it work?

Snazzie:
> Does anyone know if sharex works on a arm64 device? I know it’s not native. But will it work?

it works.

[...]

PR #7228, opened 2024-01-22
Native Arm64 support

reflectronic:
ShareX runs natively on Arm64 systems with this patch.

In .NET Framework, it is not currently possible to make an AnyCPU EXE that launches natively as Arm64. We use two approaches:

- For the Inno Setup-based installer, we set the `PreferredMachine` option under the `Image File Execution Options` key. This instructs the Windows loader to run ShareX.exe natively as Arm64. Unfortunately, since this key is under `HKEY_LOCAL_MACHINE`, this is only possible when installing as Administrator. I'm not sure if it's worth doing anything more complicated.
- For the Microsoft Store, since setting the registry key is not possible, we need to build ShareX specifically targeting Arm64. We now produce an `msixbundle` file that contains both the x64 `msix` and Arm64 `msix` inside. This file can be uploaded to the Microsoft Store just like the `appx` file used currently. The setup progam and actions workflow have been reworked to read from `win-x64` and `win-arm64` build output folders. (Note that the Inno Setup build is unchanged and uses AnyCPU as before.)

FFmpeg and Recorder-devices are both still x64 emulated. They are still both working on Arm64 though, so it can be left like this for now.

I excluded support for x86 in this patch (since I understand that you will be dropping it in the future) but it would not be difficult to add back if that's desired.

Closes https://github.com/ShareX/ShareX/issues/6653


reflectronic:
image

jimaek:
Can we get this opened and merged? Especially now with the new wave of good Windows ARM laptops

sameeraman:
Agree! can we please open this again and get it merged ?  Windows is transitioning to ARM rapidly now with the new Snapdragon xElite processors in the market

worldsdream:
+1

fizzyfrys:
Recently got a snapdragon x plus laptop, and would love to install arm64 native shareX on it.
+1

kirillrocks:
+1

rexxDZN:
+1

PR #7294, opened 2024-03-05
Translation: Hebrew missing strings

erelado:
Added missing and newly generated strings (new devbuild options), and updated my GitHub page


Jaex:
Thanks.

Issue #7073, opened 2023-09-07
Remove OCR notification


MakerManu:
Hi all, how can I remove the successful confirmation notification in the bottom right corner of the Desktop when I manually do an OCR? I know there are options to remove notifications but they work for, indeed, all capture options except for the OCR tool.
I have tried them all but it keeps appearing.
https://imgur.com/Ukkz90L

stevekenney318:
Interesting ... I've never seen that when I capture OCR ...

MakerManu:
Yes because usually Optical Character Recognition (that window with the OCR result) comes up, only I put "Process OCR silently" in the options.

MakerManu:
So can it be solved?

Eitt1:
https://www.reddit.com/r/sharex/comments/7dy7cr/how_to_remove_preview_in_bottom_right_of_screen/

It may be useful

PR #7387, opened 2024-04-10
Fix toast notification settings not being respected in OCR

Rapptz:
This should fix https://github.com/ShareX/ShareX/issues/7073

I wasn't sure if there was a code style to adhere to w.r.t. if statements being in a single line. I saw it done in certain places so I just followed that. I also admit to not testing this but the change set seemed simple enough. I edited this file using GitHub's editor (sorry) so it looks like it added an extraneous newline to the end of the file.


Rapptz:
Is there a way to get this PR open again? It's odd to have this automatically closed with zero notification or reviews of any sort.

Issue #7190, opened 2023-12-21
Cut out - black placeholder in ShareX, white in other apps


leoshortik:
Hi,
I have a problem with cut out feature. The placeholder (I hope it's correct name 😁) has black color in ShareX editor as well as in ShareX preview forms. But when I open that image in any other app like MS Paint, MS Word, web browser etc. the placeholder is white as you can see in video:

https://github.com/ShareX/ShareX/assets/111448750/9b7cf4dd-ccf6-4ddd-96b7-f5a247257858

flmbray:
I came here because of this issue also.  This appears to be because you are using the dark theme, and the cut-out feature is actually using transparency (not black).  If you change your theme to light you will see that when you perform the cut-out, the color now appears light in color (in fact to me it appears that it is exactly the same color as the Image Editor background.)  When the file is pasted into other apps, the transparency is getting lost and replaced with white.  It's not a great answer, but one workaround could be to paste into a new image in something like paint.net and then just fill the cut-out.

PR #7390, opened 2024-04-12
Add option to set background color for CutOut tool

flmbray:
Previously the CutOut tool only allowed Transparent background color.  This PR adds a "Tool Options" color picker to allow user to select the color for the background.  This resolves the issue when pasting the image into other applications forcing white as the background color.  Addresses the issue in https://github.com/ShareX/ShareX/issues/7190.


Jaex:
Thanks!

ffshurhold:
Looking forward to this, will be appreciated.

PR #7459, opened 2024-05-17
Added Registry Settings

JoshuaMaitland:
## The registry settings are DisableLanguageChange, ScreenshotsPath & UpdateChannel
Also, added an label in the General that says the settings are controlled by your IT administrator (this needs to be translated) & disabled the folder path text boxes (only if the registry settings named: PersonalPath & ScreenshotsPath exists)

ShareX_2024-05-17_16-50-40


JoshuaMaitland:
Bumping this

BrycensRanch:
Hi, just peeping in. Why would you want to disable language selection as a IT admin?

PR #7532, opened 2024-07-20
Translated missing DE strings, fixed some translations

iDazai:
Took the other pending DE translation into account.
Fixed some strings where the dash, colon were missing as in the original.
Some strings had the ampersand in it.
Translated all missing DE strings. But not those which have the same name as in the original.


Jaex:
Thanks!

PR #7561, opened 2024-08-10
Added a way to finally change the proportional fixed aspect ratio size

jtm732:
Added a way to finally change the proportional fixed aspect ratio size from 1:1 to a custom value. Previously people could only draw a square region using shift, now they can more easily select 4:3 and 16:9 images. A long awaited feature that went stale several times.


thewaliii:
But how to enable ut white holding Shift?

jtm732:
If this pull request gets committed. You will be able to change the ratio in the Options during the capture. You can set it from 1 width 1 height to something like 4 width 3 height for 4:3

PR #7562, opened 2024-08-11
Update zh-TW localization

candela97:
BTW source text related to Scrolling Capture seems to be missing.


Jaex:
Thanks I added missing forms now, you can translate them.

https://github.com/ShareX/ShareX/commit/40fc9daeccc5c06d50f0a7b482bc9cc825cf71b5

PR #7580, opened 2024-08-29
Fix image not being inserted in center when zoom level is changed

### LINKED ISSUE #7579: "Insert image in the center" is not calculated correctly when the image zoom is not 100%
1. Take a screenshot of the screen
2. Zoom out or zoom in the image in the editor
3. Insert an image from screen
4. Take a small screenshot
5. Click Insert image in the center
6. The image is not inserted in center

Without debugging, I guess the bug is here if it does not account for the zoom:

https://github.com/ShareX/ShareX/blob/6673e96f87470bff406b635c3379c00ebda11517/ShareX.ScreenCaptureLib/Shapes/Drawing/ImageDrawingShape.cs#L72

xPaw:
Fixes #7579


Jaex:
Thanks for the fix.

PR #7595, opened 2024-09-09
Fix incorrect PinToScreen toolbar visible state and #7512

### LINKED ISSUE #7512: Bug: UpdateControls called in the wrong scope
The [`UpdateControls`](https://github.com/ShareX/ShareX/blob/develop/ShareX/Tools/PinToScreen/PinToScreenForm.cs#L62) method in `PinToScreenForm#ImageScale` should be called in the same scope as [`AutoSizeForm`](https://github.com/ShareX/ShareX/blob/develop/ShareX/Tools/PinToScreen/PinToScreenForm.cs#L59), it'll cause a crash if the initial scale changed. Since `UpdateControls` will be called [right before loaded](https://github.com/ShareX/ShareX/blob/develop/ShareX/Tools/PinToScreen/PinToScreenForm.cs#L157), this modification should be fine.

I found this bug when trying to add a custom command to pin images with CLI. This is the code if you're interested:
<details>
<summary>Click to expand</summary>
<pre><code class = "language-csharp language-c#">// i did it in dnSpy because i don't want to setup a compiler
// ShareX 16.1
private bool CheckPinCmd(CLICommand command)
{
  if (!command.CheckCommand("PinCmd")) return false;
  try {
    // path, x, y, scale
    string[] par = command.Parameter.Split(',');
    if (par.Length != 4) return true;
    string path = par[0];
    int x = int.Parse(par[1]);
    int y = int.Parse(par[2]);
    int scale = int.Parse(par[3]);
    Image image = ImageHelpers.LoadImage(path);
    PinToScreenOptions options = new PinToScreenOptions();
    options.InitialScale = scale;
    options.Placement = ContentAlignment.TopLeft;
    options.BorderSize = 1;
    options.Shadow = false;
    TaskHelpers.PinToScreen(image, new Point(x, y), options);
  } catch {}
  return true;
}</code></pre>
</details>

aMelonRind:
When two PinToScreen images has overlap and moving mouse from the bottom one to the top one, the toolbar on the bottom one stays, which is unexpected behavior.

Also fixes #7512 since it's 1 line and opened by me.

### DIFF
--- a/ShareX/Tools/PinToScreen/PinToScreenForm.cs
+++ b/ShareX/Tools/PinToScreen/PinToScreenForm.cs
@@ -125,7 +125,7 @@ namespace ShareX
 
         public PinToScreenOptions Options { get; private set; }
 
-        private bool isDWMEnabled, loaded, dragging;
+        private bool isDWMEnabled, loaded, dragging, imageHovered;
         private Point initialLocation;
         private Cursor openHandCursor, closedHandCursor;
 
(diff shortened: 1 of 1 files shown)


aMelonRind:
seems like commits pushed from web editor have terrible diff.

PR #7620, opened 2024-10-01
Update to .NET 8

benbryant0:
I tried updating to .NET 8 and fixed what was broken. Very little needed to be fixed to get it working, but a bunch of new warnings showed up so I fixed the ones that were trivial.

Since there aren't any tests in code and I only just stumbled onto this tool, I haven't exactly tested it thoroughly. I opened every window/section in the UI and poked around for a while and made sure every test under the debug menu worked.

If you don't want to go to .NET 8 yet, every commit before `Update to .NET 8` should be completely fine for .NET Framework as well.

The `ImageListView` dependency is still on .NET Framework and has some shortcomings that cause problems with the optimisations to LINQ in modern .NET. There's an [old PR there](https://github.com/oozcitak/imagelistview/pull/234) to fix it, but I've worked around the problem for now.

It still won't be possible to use the `dotnet` CLI tool for building (and maybe other things) due to the COM reference in `ShareX.HelpersLib` (used for creating & checking shortcuts). I'm pretty sure that could be replaced with something more modern, but I've never dealt with that stuff and didn't want to risk messing it up.

### DIFF
--- a/ShareX/ShareX.csproj
+++ b/ShareX/ShareX.csproj
@@ -1,17 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net48</TargetFramework>
+    <TargetFramework>net80-windows10.0.26100.0</TargetFramework>
     <OutputType>WinExe</OutputType>
     <ApplicationIcon>ShareX_Icon.ico</ApplicationIcon>
     <UseWindowsForms>true</UseWindowsForms>
     <ApplicationManifest>app.manifest</ApplicationManifest>
     <Description>Screen capture, file sharing and productivity tool</Description>
   </PropertyGroup>
-  <ItemGroup>
-    <Reference Include="System.Design" />
-    <Reference Include="System.IO.Compression" />
-    <Reference Include="System.Web" />
-  </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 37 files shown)


ShareX/ShareX.csproj
```diff
@@ -1,17 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net48</TargetFramework>
+    <TargetFramework>net80-windows10.0.26100.0</TargetFramework>
```
Vertygo, line 3: Not sure if its intentional but in other csproj files you had `net80-windows`

benbryant0: Hah, yeah.. I never remember to write explanations in my commits 🤦‍♂️

I didn't know about this before, but this is the 'replacement' of the reference to `Microsoft.Windows.SDK.Contracts`. It specifies what version of the Windows APIs are available (used [here](https://github.com/ShareX/ShareX/blob/ec3ce9caebb4ac5a44d7e95e76f1e8f95cae049f/ShareX/Tools/OCR/OCRHelper.cs#L35-L38) for example). [Here's](https://learn.microsoft.com/en-us/dotnet/standard/frameworks#os-version-in-tfms) a section about this in the .NET docs.

peymanr34:
I'm working on a fork of ImageListView (pending release). If anyone's interested to look around you can find it here: https://github.com/imagelistview-core/imagelistview-core

Issue #6799, opened 2023-03-08
Display keypresses while recording GIF


governos:
Hello, 

I do volunteer support and I often use the GIF recording to show steps to users.
I think it would be useful to display the pressed keys/mouse clicks on the GIF so the exact steps could be shown.

Thank you.

weitzhandler:
Related: 

#5570 #4097 #6785

Issue #6785, opened 2023-02-28
When recording a video or GIF, the mouse click effect


ghost:
Hope to add an option，When recording GIF, add a click effect when clicking the mouse，So that others know when they click on the screen。

weitzhandler:
I see this has been reported before.
#5570, #4097

Related: #6799

Issue #5570, opened 2021-05-14
Suggestion: Capture clicks when recording GIFs


alejandro5042:
Add an option that captures clicks in GIFs with a halo, similar to:

![image](https://user-images.githubusercontent.com/3143819/56770600-063f6600-677a-11e9-84e9-fcbdfd985444.png)

(Previous issue: #4097)

Issue #4097, opened 2019-04-25
Suggestion: Capture clicks when recording GIFs


alejandro5042:
Add an option that captures clicks in GIFs with a halo, similar to:

![image](https://user-images.githubusercontent.com/3143819/56770600-063f6600-677a-11e9-84e9-fcbdfd985444.png)

MikelLuri:
I'd love to have this, too

alejandro5042:
I still want this, thanks!

alejandro5042:
I still care about this. Would make a big positive impact to recording issues to post in bug trackers.

alejandro5042:
Darn it, I still care about this. Missed the last message! Can we reopen?

PR #7622, opened 2024-10-02
Add mouse click effect capability for screen recording

Vertygo:
Added a mouse click effect feature for screen recording. This will enable people to see when click is performed in video.

This PR covers the following feature requests
#6785
#5570
#4097
#6799

Example of feature in action

https://github.com/user-attachments/assets/54db845f-b5fd-448d-8e63-dc7979c226ba

Review comments todo:

- [x] run effect (form) outside Task.Run to avoid a need for explicit execution on main UI thread (Form.Invoke(...))
- [x] don't use underscore variables
- [x] right button support
- [ ] add support for customizable color, size, fade time (make it as standalone tool)
- [x] format code
- [ ] fade animation
- [x] fix rider changes in settings form

### DIFF
--- a/ShareX.HelpersLib/Input/MouseClickEffectForm.cs
+++ b/ShareX.HelpersLib/Input/MouseClickEffectForm.cs
@@ -1,24 +1,23 @@
 ﻿using System;
 using System.Drawing;
+using System.Drawing.Drawing2D;
 using System.Windows.Forms;
 
 namespace ShareX.HelpersLib
 {
-    // transparent form where mouse effect will be drawn
-    public class MouseClickEffectForm : Form
+    // transparent form where mouse click effect will be drawn
+    public class MouseClickEffectForm : LayeredForm
     {
-        private bool _drawEffect = false;
-        public MouseClickEffectForm()
+
+        protected override CreateParams CreateParams
         {
-            Size = new Size(40, 40);
-            FormBorderStyle = FormBorderStyle.None;
-            ShowIcon = false;
-            ShowInTaskbar = false;
-            TopMost = true;
-            AllowTransparency = true;
-            BackColor = Color.White; // Set a key color for transparency
-            TransparencyKey = Color.White; // Make that color transparent
-            Opacity = 0.3;
+            get
+            {
+                // set style to layered topmost transparent
+                var createParams = base.CreateParams;
+                createParams.ExStyle |= (int)(WindowStyles.WS_EX_TOPMOST | WindowStyles.WS_EX_TRANSPARENT);
+                return createParams;
+            }
         }
 
         /// <summary>
@@ -26,9 +25,8 @@ namespace ShareX.HelpersLib
         /// </summary>
         public void DrawMouseEffect(Point cursorPosition)
         {
-            _drawEffect = true;
             CenterFormToCursorPosition(cursorPosition);
-            Invalidate(); // redraw form
+            SelectBitmap(GetCircleImage(Color.Red), 100);
         }
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


ShareX.HelpersLib/Input/MouseClickEffectForm.cs
```diff
@@ -0,0 +1,79 @@
...
+                // Define a pen to draw hollow circle
+                Pen pen = new Pen(Color.Red, 3);
+                int diameter = 10;
+
+                // Calculate the top-left corner to center the circle
+                int x = (ClientSize.Width - diameter) / 2;
+                int y = (ClientSize.Height - diameter) / 2;
+
+                // .NET GDI+ is not precise when drawign circles this would be better off with WPF/vector-based drawing
+                g.DrawEllipse(pen, x, y, diameter, diameter);
```
Vertygo, line 60: This might not be best solution due to GDI+ poor precision.
ShareX_CwfyI1M3JM

nicknijenhuis:
Design wise, something like this would be great. 

https://www.vecteezy.com/video/8079876-pointer-hand-cursor-clicking-technology-and-internet-icons-animation-on-green-screen-background-mouse-click-symbol-with-spark-on-green-screen

But let me not interfere with this to much! Your PR looks already very nice without my comment. 

Thanks, have been wanting this!

ShareX.HelpersLib/Input/MouseClickEffectForm.cs
saintcore: I can't add that much to this, I just wanted to say that rn I'm using this tool https://github.com/Phaiax/Key-n-Stroke to achieve the same and as it is open-source already you guys can probably get inspiration from there and/or took over some of the code to not reinvent the wheel here :) (altho I dunno if the poor precision thing is really better with keynstroke --> at least from a enduser-experience I've never had simliar problems)

Vertygo: Good point @saintcore. I do get inspired by projects like this. I like keyboard capturing capability it has and I might take a stab at it after I'm done with mouse effect.

I don't know how things work regarding "taking over" some code in the open-source world. I guess some collaboration with the owner (@Phaiax) is needed.

Regarding the small issue I mentioned here, I think that might be solved with help from @Jaex but keynstroke uses WPF so it is slightly different than ShareX which uses Windows Forms.

Vertygo:
@nicknijenhuis a nice effect, I think this could be achieved with a combination of timer(s) and GDI+. If there is enough interest for it, I could spend some time implementing it at later point in separate PR as I would like to avoid scope creeping this PR otherwise it won't be finished soon.

ShareX.HelpersLib/Input/MouseClickEffectForm.cs
saintcore: Well according to the provided license, it shouldn't be a problem (see https://github.com/Phaiax/Key-n-Stroke?tab=Apache-2.0-1-ov-file#readme ). A good summary of what apache 2.0 license allows us to do with and not is available over at snyk: https://snyk.io/learn/apache-license/

> Under the Apache license, users are permitted to modify, distribute, and sublicense the original open source code. Commercial use, warranties, and patent claims are also allowed with the Apache license. 
....
When using the Apache license, developers must include the original copyright notice, a copy of the license text itself, and in some cases, a copy of the notice file with attribution notes and a disclosure of any significant changes made to the original code.
....

With that said, ofc it would be nice to hear from the original author of keynstroke and his desires beside license thoughts :)

PR #7664, opened 2024-11-05
Capture delay french translation fix

theiereman:
TrayScreenshotDelay3 had "2 seconds" as value instead of "3 seconds" in french translation


theiereman:
Moving that fix into that pull request : https://github.com/ShareX/ShareX/pull/7663

PR #7673, opened 2024-11-13
fix/7381 updating ImageShack urls to use correct format and https

### LINKED ISSUE #7381: Suggestion - Option to enable Direct Link for Imageshack subscribers
When I use Imageshack as an ImageUploader and paste the URL, I get a link like this:

http://imageshack.com/a/img923/746/jI3tjc.png

I would like the option to get the direct link instead. 

https://imagizer.imageshack.com/a/img923/746/jI3tjc.png

You can see that the first URL redirects to the direct link. 

I have a paid ImageShack subscription and have access to my direct link. See here: http://imageshack.com/a/img922/8948/TrHsBn.png 

The reason I ask is that since imgur stopped direct linking, I needed an alternative. The app I paste my URL in does not allow redirect and I end up having to copy my links from the webpage. Which isn't very productive. It would save me some time to have the direct link copied to clipboard instead. 

Thank you!

joejoe04:
closes #7381 

URLs for images uploaded to ImageShack are currently poorly formatted to use `http` and the incorrect domain (`imageshack.com` rather than `imagizer.imageshack.com`), which results in unnecessary rewrites.

This fixes that so the URLs are formatted such that no rewrites are required.

### DIFF
--- a/ShareX.UploadersLib/ImageUploaders/ImageShackUploader.cs
+++ b/ShareX.UploadersLib/ImageUploaders/ImageShackUploader.cs
@@ -118,8 +118,8 @@ namespace ShareX.UploadersLib.ImageUploaders
                         if (uploadResult != null && uploadResult.images.Count > 0)
                         {
                             ImageShackImage image = uploadResult.images[0];
-                            result.URL = string.Format("http://imageshack.com/a/img{0}/{1}/{2}", image.server, image.bucket, image.filename);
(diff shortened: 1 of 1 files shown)


LuckyPadddy:
This addition makes me very happy!

LuckyPadddy:
@joejoe04 Thanks for letting me know!

PR #7706, opened 2024-12-06
fixed #5378: Recognize Jira rebrand

### LINKED ISSUE #5377: Do not force user to accept GPL but just show it
I as end-user am not affected by the GPL.
It only applies to modifying and distributing.
I should not be forced to formally accept it.
Instead it should just be shown with a "Next" button
in the installer instead of being forced to accept it.

### LINKED ISSUE #5378: JIRA -> Jira
Since quite some time JIRA did rebrand to Jira.
The ShareX UI still uses JIRA currently.

BrycensRanch:
fixed #5378


BrycensRanch:
Apologies - I had the wrong issue number in my clipboard :laughing:

PR #7846, opened 2025-03-15
Add ClearClipboardAfterCapture option for screenshots

NebulaMods:
Introduce a new boolean property `ClearClipboardAfterCapture` in `ApplicationConfig.cs` and `TaskSettings.cs` to allow users to clear the clipboard after capturing a screenshot. Update `CaptureInternal` in `CaptureBase.cs` to implement this functionality.


jhudler:
It _might_ also address #7909

PR #7887, opened 2025-04-17
Disable upload by default

### LINKED ISSUE #7535: Disable Upload By Default
Hi, as a user who just wanted to use the convenient screen capture feature of ShareX, it took me by surprise when I found out my captured image was uploaded to imgur automatically. It is giving me a big headache because the image contains private info, and imgur is ignoring my removal request... :dizzy_face::dizzy_face::dizzy_face:

Please consider disabling the image upload function by default. Even if image upload is an important feature of this app, it would be much better if the option to enable automatic upload is prompted on first startup (but with warning messages).

Related past issues:

- #4719
- #3403

lirc572:
Resolves #7535.

This PR simply makes `Program.Settings.DisableUpload` default to true.

If this is not desired, let's discuss about alternative solutions in #7535.

### DIFF
--- a/ShareX/ApplicationConfig.cs
+++ b/ShareX/ApplicationConfig.cs
@@ -236,7 +236,7 @@ namespace ShareX
         [Category("Image"), DefaultValue(false), Description("Strip color space information chunks from PNG image.")]
         public bool PNGStripColorSpaceInformation { get; set; }
 
-        [Category("Upload"), DefaultValue(false), Description("Can be used to disable uploading application wide.")]
+        [Category("Upload"), DefaultValue(true), Description("Can be used to disable uploading application wide.")]
         public bool DisableUpload { get; set; }
 
         [Category("Upload"), DefaultValue(false), Description("Accept invalid SSL certificates when uploading.")]


Xenthys:
An alternative solution is already implemented, by showing a warning on first upload that explains it'll send your screenshot to Imgur and forcing the user to wait 5 seconds before they can accept. They can refuse immediately though, which will disable uploads until manually reinstated by the user in question.

Since most users still use **Share**X to **share** screenshots, disabling uploads by default and forcing them to dive into advanced settings isn't desirable, unless you are ready to provide a dedicated support team to the ShareX project to handle the influx of questions and complaints from old and new users alike.

lirc572:
> An alternative solution is already implemented, by showing a warning on first upload that explains it'll send your screenshot to Imgur and forcing the user to wait 5 seconds before they can accept. They can refuse immediately though, which will disable uploads until manually reinstated by the user in question.

I remember not seeing this warning when I experienced the problem. I'll re-install the latest version to see if it works as intended. If it does, it sure solves half of the problem. I still personally believe it's better to default to not uploading pictures. E.g. if after taking the first screenshot, the user is shown a success screen with a checkbox for uploading to imgur (with warning sign and unchecked by default), that would be more sensible to me, requiring just one more click for users who like auto-upload while preventing others from accidentally enabling the feature.

> Since most users still use ShareX to share screenshots, disabling uploads by default and forcing them to dive into advanced settings isn't desirable

I agree that we should not disable auto-upload for existing users enjoying the feature who upgrade to a version that makes auto-upload disabled by default.

Nerdoholic:
> An alternative solution is already implemented.

Please read https://github.com/ShareX/ShareX/issues/7535 as to why this is not the proper way to deal with this.

Otherwise I 100% agree with you. Changing this is necessary, yet it should be done in a sane way. Only swapping the Default value will bring much frustration to the existing userbase/nooby newcomers that simply want to use it that way.

E.g. improving the FirstLaunch wizard to enable exactly that userbase to have a great experience using the app is very desirable.

Not kicking the old users in the nuts is doable as well, simply implement a migration for the variables of older versions (there's some finetuning to do here of course regarding user action templates, but it's not that complicated).

But currently the project is frightening professional environments with that setting, and that's simply an unnecessarily dumb thing to do.

2mustange:
This is great and follows better security practices

2mustange:
A bit disappointing this is considered "stale" without devs discussing it

Nerdoholic:
> Closed with unmerged commits
> 
> This pull request is closed.
What a shitshow of wannabe grownups too incompetent to handle an obviously problematic design even if pointed out properly.

Pathetic.

PR #7926, opened 2025-05-11
Add the option to copy the normalized RGBA colors in the color picker

AregevDev:
Hello,
I use the color picker tool frequently, I primarily work on graphics related applications that usually accept normalized colors, between 0 and 1. It would be great to have this functionality in ShareX, it would help my workflow quite a bit (having the colors as constants instead of dividing by 255 every time)

I implemented another `ToolStripMenuItem` in the color picker's `Copy` button. I followed the conventions of the other copy menu items. I formatted the output to include a single zero after the dot in case the number is whole (TIL C# omits all zeros even in floating point types)

The only thing left is to localize the new string I added for the button text.
I hope it is accepable


Jaex:
Did you check this?

AregevDev:
Oh what? What do you know

PR #7976, opened 2025-06-13
Clip a child window against its parent

### LINKED ISSUE #7947: Window capture should take parent region into account
When taking a screenshot of a window,
ShareX does not take the region of the parent window into account:

Image

Here an example with the default Visual studio win32 project is edited so that the ok button extends wider than the dialog.
ShareX should clip the ok button to the dialog, not allow it to be screenshotted while the mouse is outside the parent window.

The 'question behind the question' is:
On Win11 there is no native support for a vertical taskbar anymore.
[Windhawk](https://windhawk.net/) with the [Vertical Taskbar for Windows 11](https://windhawk.net/mods/taskbar-vertical) allows to work around this, but the only way to get it working is to tell one of the child windows of the taskbar that it actually occupies a bigger area than it does. (The top level window of the taskbar is smaller.)

Trying to take a screenshot from a window is quite hard on a multi-monitor setup, because a big part of the screen is occupied by the child window that is 'too big':

Image

learn-more:
This fixes #7947

### DIFF
--- a/ShareX.ScreenCaptureLib/Helpers/WindowsRectangleList.cs
+++ b/ShareX.ScreenCaptureLib/Helpers/WindowsRectangleList.cs
@@ -53,9 +53,12 @@ namespace ShareX.ScreenCaptureLib
                     cts = new CancellationTokenSource();
                     cts.CancelAfter(Timeout);
                 }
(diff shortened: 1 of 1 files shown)


Jaex:
Thanks.

Issue #8047, opened 2025-08-12
unable to open database file 18.0 pre-release


xlorm:
Windows 11 23H2 (22631.5624)

Updated to the pre-release version and now ShareX crashes every time it tries to start with the below error.

code = CantOpen (14), message = System.Data.SQLite.SQLiteException (0x87AF03F3): unable to open database file
   at System.Data.SQLite.SQLite3.Open(String strFilename, String vfsName, SQLiteConnectionFlags connectionFlags, SQLiteOpenFlagsEnum openFlags, Int32 maxPoolSize, Boolean usePool)
   at System.Data.SQLite.SQLite3.Open(String strFilename, String vfsName, SQLiteConnectionFlags connectionFlags, SQLiteOpenFlagsEnum openFlags, Int32 maxPoolSize, Boolean usePool)
   at System.Data.SQLite.SQLiteConnection.Open()
   at ShareX.HistoryLib.HistoryManagerSQLite.Connect(String filePath)
   at ShareX.HistoryLib.HistoryManagerSQLite..ctor(String filePath)
   at ShareX.SettingManager.HistoryConnect()
   at ShareX.SettingManager.LoadApplicationConfig(Boolean fallbackSupport)
   at ShareX.SettingManager.LoadInitialSettings()
   at ShareX.Program.Run()
   at ShareX.Program.Main(String[] args)

ShareX 18.0

ct-rivd:
Encountering the same issue.
It seems to be occurring for everyone who is logging on to a remote desktop server where ShareX was upgraded from 17.1 to 18.0

First run after upgrade
```
ShareX 18.0
2025-08-14 07:54:56.673 - ShareX continue.
2025-08-14 07:55:24.736 - ShareX starting.
2025-08-14 07:55:24.738 - Version: 18.0
2025-08-14 07:55:24.739 - Build: Release
2025-08-14 07:55:24.739 - Command line: "C:\Program Files\ShareX\ShareX.dll"
2025-08-14 07:55:24.739 - Personal path: \\FILESERVER\MyDocuments$\username\My Documents\ShareX
2025-08-14 07:55:24.740 - Operating system: Windows Server 2022 Standard (64-bit)
2025-08-14 07:55:24.740 - .NET version: 9.0.8
2025-08-14 07:55:24.740 - Running as elevated process: False
[...]

PR #8052, opened 2025-08-14
Restore UNC path support by parsing connection string with .NET

BrycensRanch:
Previously, opening a SQLite database stored on a network share via a UNC path

 e.g. `\\server\share\file.db`) would fail with `unable to open database file`

 This change uses .NET’s connection string parsing to correctly handle UNC paths,
 ensuring databases on network shares can be opened without errors.

The previous failure of parsing lies in `System.Data.Sqlite`. SnapX does not have this bug because it uses Microsoft's `Microsoft.Data.Sqlite`, which has never had this exact bug, while the bug has [stayed open for over a decade in `System.Data.Sqlite`](https://system.data.sqlite.org/home/info/283344397b)

If ShareX runs into any further issues with `System.Data.Sqlite`, I'm happy to send a PR to switch it over.

This change has been tested on SMB3 shares and works on both network drive letters and UNC paths. Most importantly, it has been tested against regular paths. 

Fixes ShareX/ShareX#8047

### DIFF
--- a/ShareX.HistoryLib/HistoryManagerSQLite.cs
+++ b/ShareX.HistoryLib/HistoryManagerSQLite.cs
@@ -28,7 +28,6 @@ using ShareX.HelpersLib;
 using System;
 using System.Collections.Generic;
 using System.Data.SQLite;
-using System.Text.RegularExpressions;
 
 namespace ShareX.HistoryLib
 {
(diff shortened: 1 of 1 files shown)


Jaex:
Thanks!

PR #8104, opened 2025-08-26
Add EnableConnectionRelease option

### LINKED ISSUE #8061: Compatibility issue with OneDrive
Hello, today I have noticed that my Sharex folder is not syncing into OneDrive. 
For some reason the OneDrive blocks the file "History.db" from being uploaded.

Image

Image

Can you please consider re-naming this file to allow Sharex folder to be synced into OneDrive?

**Workaround**
Workaround for this issue is to downgrade to 17.0.0

BrycensRanch:
With this setting, affected users will have to manually enable it in the advanced settings 

Fixes #8061

### DIFF
--- a/ShareX.HistoryLib/HistoryManagerSQLite.cs
+++ b/ShareX.HistoryLib/HistoryManagerSQLite.cs
@@ -23,20 +23,22 @@
 
 #endregion License Information (GPL v3)
 
+using Microsoft.Data.Sqlite;
 using Newtonsoft.Json;
 using ShareX.HelpersLib;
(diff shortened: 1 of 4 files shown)


BrycensRanch:
> > No longer needed by users, it seems.
> 
> We need! Is this approach allowing an open and close connection for each operation? I'm no developer and have no experience using Github fyi. I defer to your expertise. Thanks for your efforts building this.


Yes, it opens and then immediately closes the connection once the operation is done. I will add one needed fix that will make it thread-safe.

Issue #7983, opened 2025-06-19
[FEATURE REQUEST] Add a command prompt flag that automatically starts autocapture


jfaquinojr:
There is already a "-AutoCapture" flag, but what it does is it just shows the AutoCapture form. 

What I need is something that will actually start the autocapture as if I clicked the Execute button.

Something along the lines of:

```shell
sharex -AutoCapture -Start 10s
```

This would start the auto capture with interval of 10 seconds with the default capture settings (or pulled from a config file).

Please let me know if this would be something that can be merged later, because I'd be happy to work on it.

Jaex:
Did you try: `-StartAutoCapture`

jfaquinojr:
that worked, thanks!! Before I close this, I didnt see anything for -StopAutoCature. Is there a way to do this via command line?

PR #8156, opened 2025-09-21
Implemented issue #7983 third comment request

BallisticLingonberries:
> I didnt see anything for -StopAutoCature. Is there a way to do this via command line?

Added HotKey and therefore Command Line support to stop auto capture without opening the auto capture window. I used the same clock icon for the hotkey icon resource, but this can be changed if need be. Let me know if I need to edit anything.

Thanks Jaex.


Jaex:
Thanks.

BallisticLingonberries:
Very welcome. Thanks again for the great software.

Issue #7461, opened 2024-05-17
Error on startup - Unable to register hotkey


Skorki51:
Hello when I try to run sharex I get an error message every time.

![Zrzut ekranu 2024-05-17 203341](https://github.com/ShareX/ShareX/assets/149414795/67036fa9-8b3b-4c5c-b3e5-2dcb5a3f8479)

Here I add a debug log from the program startup:

> 2024-05-17 20:41:37.828 - ShareX starting.
2024-05-17 20:41:37.829 - Version: 16.1
2024-05-17 20:41:37.829 - Build: Steam
2024-05-17 20:41:37.829 - Command line: "D:\STEAM\steamapps\common\ShareX\ShareX\ShareX.exe" 
2024-05-17 20:41:37.829 - Personal path: D:\OneDrive\Dokumenty\ShareX
2024-05-17 20:41:37.829 - Operating system: Windows 10 Pro (64-bit)
2024-05-17 20:41:37.830 - Running as elevated process: False
2024-05-17 20:41:37.844 - ApplicationConfig load started: D:\OneDrive\Dokumenty\ShareX\ApplicationConfig.json
2024-05-17 20:41:38.045 - ApplicationConfig load finished: D:\OneDrive\Dokumenty\ShareX\ApplicationConfig.json
2024-05-17 20:41:38.050 - UploadersConfig load started: D:\OneDrive\Dokumenty\ShareX\UploadersConfig.json
2024-05-17 20:41:38.054 - Language changed to: English (United States)
2024-05-17 20:41:38.055 - MainForm init started.
2024-05-17 20:41:38.132 - UploadersConfig load finished: D:\OneDrive\Dokumenty\ShareX\UploadersConfig.json
2024-05-17 20:41:38.133 - HotkeysConfig load started: D:\OneDrive\Dokumenty\ShareX\HotkeysConfig.json
2024-05-17 20:41:38.155 - HotkeysConfig load finished: D:\OneDrive\Dokumenty\ShareX\HotkeysConfig.json
2024-05-17 20:41:38.363 - MainForm init finished.
2024-05-17 20:41:38.619 - Hotkey registered: Hotkey: Ctrl + Print Screen, Description: Capture region, Job: RectangleRegion
2024-05-17 20:41:38.619 - Hotkey registered: Hotkey: Print Screen, Description: Capture entire screen, Job: PrintScreen
2024-05-17 20:41:38.620 - Unable to register hotkey: Alt + Print Screen
[...]

Issue #6581, opened 2022-11-07
OneDrive causing hotkey conflict


mbressman:
Received the below error when I recently rebooted my PC (after setting up OneDrive personal).  I was able to get around it by simply closing out of ShareX and OneDrive and then re-launching ShareX first and then OneDrive, but how do I avoid this in the future?  Thanks!

![ShareX - Hotkey registration failed](https://user-images.githubusercontent.com/7513711/200228629-a2b00f53-8bf6-4615-93ac-9315c35384cf.png)

FrankE-Aqura:
Received the same error. I'm a new user so it wasn't the best thing to see the first time I started the program.
Running Windows 11 21H2 (22000.1098), ShareX 14.1 and OneDrive Personal and OneDrive for Business (22.131.0619.0001 64-bit)

> 2022-11-16 08:47:45.663 - Loading new HotkeysConfig instance.
> 2022-11-16 08:47:45.666 - Language changed to: English (United States)
> 2022-11-16 08:47:45.670 - MainForm init started.
> 2022-11-16 08:47:46.078 - MainForm init finished.
> 2022-11-16 08:47:46.250 - Startup time: 2125 ms
> 2022-11-16 08:47:46.271 - Hotkey registered: Hotkey: Ctrl + Print Screen, Description: Capture region, Job: RectangleRegion
> 2022-11-16 08:47:46.272 - Unable to register hotkey: Print Screen
> 2022-11-16 08:47:46.272 - Hotkey register failed: Hotkey: Print Screen, Description: Capture entire screen, Job: PrintScreen
> 2022-11-16 08:47:46.272 - Unable to register hotkey: Alt + Print Screen
> 2022-11-16 08:47:46.272 - Hotkey register failed: Hotkey: Alt + Print Screen, Description: Capture active window, Job: ActiveWindow
> 2022-11-16 08:47:46.272 - Hotkey registered: Hotkey: Shift + Print Screen, Description: Start/Stop screen recording, Job: ScreenRecorder
> 2022-11-16 08:47:46.272 - Hotkey registered: Hotkey: Ctrl + Shift + Print Screen, Description: Start/Stop screen recording (GIF), Job: ScreenRecorderGIF
> 2022-11-16 08:53:19.529 - HotkeyManager started.
> 2022-11-16 08:53:19.530 - WatchFolderManager started.

mcouillard:
[...]

Issue #6193, opened 2022-04-04
Suggestion: Specific instructions regarding OneDrive and hotkey conflict.


martinsstuff:
I think it could be useful if the message warning about being unable to register hotkeys would instruct the user to press "Print Screen" in order to open the OneDrive dialog. On a fresh OneDrive install it will show "Automatically save screenshots I capture to OneDrive" as disabled, even though it claims the hotkey. In order to fix this the user first needs to press the hotkey and decline in the dialog.

Jaex:
ShareX can't know which application claim the hotkey.

martinsstuff:
It actually does, I forgot to take a screenshot but you can see the identical message here: https://github.com/ShareX/ShareX/issues/6051#issuecomment-1022344543

Jaex:
It guessing by scanning active apps. There is big difference.

Issue #5720, opened 2021-07-27
Conflict with OneDrive


lwchkg:
Today ShareX failed to register the shortcut keys because of conflict with OneDrive. This is not a bug of ShareX.

OneDrive registered PrtScn key without my permission. Once I pressed the PrtScn key it prompted me to use it's upload feature. I declined. Then restarted ShareX, and the shortcut keys are working normally again.

While there is no bug to fix, I am writing to notify the authors because casual users might think that it was indeed a bug in ShareX. You may want to do something to address this.

Windows version: 21H1

PR #8329, opened 2026-01-18
Add automatic retry mechanism for hotkey registration

tomstetson:
## Summary
- Add automatic retry mechanism for hotkey registration to handle race conditions at startup
- Fixes the common issue where OneDrive (or other apps) temporarily claim PrintScreen hotkeys before ShareX can register them
- New configurable settings: `HotkeyRetryEnabled`, `HotkeyRetryCount`, `HotkeyRetryDelayMs`
- Backward compatible - existing `UpdateHotkeys()` method preserved

## Problem

When ShareX starts, it immediately attempts to register global hotkeys using the Win32 `RegisterHotKey` API. Other applications like OneDrive often claim these same hotkeys (PrintScreen, Alt+PrintScreen) during their own startup sequence, causing:

1. ShareX calls `RegisterAllHotkeys()` immediately at startup
2. OneDrive has already registered the hotkey
3. `RegisterHotKey` returns `false`, hotkey status set to `Failed`
4. "Hotkey registration failed" error dialog shown
5. User must manually restart ShareX or open Hotkey Settings

**Note:** OneDrive claims these hotkeys even when its screenshot feature is disabled.

### New Method: `UpdateHotkeysWithBackgroundRetry()`
- Performs initial registration synchronously (fast startup)
- If hotkeys fail, starts background retry task (fire-and-forget)
- **Does not block `IsReady`** - CLI invocations from second instances work immediately
- Shows error dialog only after all retry attempts are exhausted

### New Configuration Settings (ApplicationConfig)

| Setting | Default | Range | Description |
|---------|---------|-------|-------------|
| `HotkeyRetryEnabled` | true | - | Enable/disable automatic retry |
| `HotkeyRetryCount` | 3 | 1-10 | Number of retry attempts |
| `HotkeyRetryDelayMs` | 2000 | 500-30000 | Delay between retries (ms) |

## Security Considerations

Settings are clamped to safe ranges to prevent DoS via malicious config files:
- Max retry count: 10 (prevents infinite loops)
- Max delay: 30 seconds (prevents extreme startup delays)
- Worst case: 10 × 30s = 5 minutes background retry window

## Files Changed

| File | Changes |
|------|---------|
| `ShareX/ApplicationConfig.cs` | Added 3 new settings for retry configuration |
| `ShareX/HotkeyManager.cs` | Added `UpdateHotkeysWithBackgroundRetry()`, `RetryFailedHotkeysInBackgroundAsync()`, and `HasFailedHotkeys()` |
| `ShareX/Forms/MainForm.cs` | Updated `InitHotkeys()` to use non-blocking retry method |

## Related Issues

- #6581 - OneDrive causing hotkey conflict
- #3431 - Hotkey registration failed (every Windows 10 install)
- #5720 - Conflict with OneDrive
- #6193 - Suggestion for OneDrive instructions
- #1144 - Windows 10 standard shortcuts fail
- #7461 - Error on startup - Unable to register hotkey

### DIFF
--- a/ShareX/ApplicationConfig.cs
+++ b/ShareX/ApplicationConfig.cs
@@ -72,6 +72,12 @@ namespace ShareX
         public HotkeyType TrayLeftDoubleClickAction = HotkeyType.OpenMainWindow;
         public HotkeyType TrayMiddleClickAction = HotkeyType.ClipboardUploadWithContentViewer;
 
+        // Hotkey retry settings - helps with race conditions at startup (e.g., OneDrive conflict)
+        // Values are clamped at runtime: HotkeyRetryCount [1-10], HotkeyRetryDelayMs [500-30000]
+        public bool HotkeyRetryEnabled = true;
+        public int HotkeyRetryCount = 3;
+        public int HotkeyRetryDelayMs = 2000;
+
         public bool AutoCheckUpdate = true;
         public UpdateChannel UpdateChannel = UpdateChannel.Release;
         // TEMP: For backward compatibility
--- a/ShareX/Forms/MainForm.cs
+++ b/ShareX/Forms/MainForm.cs
@@ -365,7 +365,9 @@ namespace ShareX
                 Program.HotkeyManager.HotkeyTrigger += HandleHotkeys;
             }
 
-            Program.HotkeyManager.UpdateHotkeys(Program.HotkeysConfig.Hotkeys, !Program.IgnoreHotkeyWarning);
+            // Use retry mechanism to handle race conditions with other apps (e.g., OneDrive)
+            // Retries run in background to avoid blocking startup and WaitFormLoad timeout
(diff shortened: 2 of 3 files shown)


tomstetson:
## Addressed Codex Review Feedback

Thanks for the review! I've updated the implementation to address the **P2 - Avoid blocking InitHotkeys on retry delays** feedback.

### Changes Made:
- **Renamed** `UpdateHotkeysWithRetryAsync()` → `UpdateHotkeysWithBackgroundRetry()`
- **Initial registration is now synchronous** - returns immediately after first attempt
- **Retries run in background** via fire-and-forget (`_ = RetryFailedHotkeysInBackgroundAsync()`)
- **`IsReady` is set immediately** after initial registration, before retries complete

### Result:
- `WaitFormLoad(5000)` now succeeds immediately since `IsReady` is set right away
- CLI invocations from second instances work correctly even during retry window
- Retries still happen in background and show error dialog only after exhausting all attempts

PR #8344, opened 2026-01-23
[Feature] PrivateBin text uploader

SoNiice:
Added PrivateBin support after cleaning up my uploading logic since I wanted to have everything separated properly. Only tested against private hosted [Docker](https://hub.docker.com/r/privatebin/nginx-fpm-alpine/) on my Unraid system. Some localization is missing, only added default (English) and German.


Jaex:
Thanks!

PR #7866, opened 2025-03-29
Arm64 support

ahossny:
Adds Support to ARM64
Migrated to :NET8

I have started this project and my aim to add arm64 support, as it would be more efficient and would not drain my battery. 

image

PR #8370, opened 2026-01-31
Arm64 support

ahossny:
Merge to resolves conflicts with version 19.0.2

Previous stale PR: [7866](https://github.com/ShareX/ShareX/pull/7866)

Builds found  [here](https://github.com/ahossny/ShareX/releases)


BrycensRanch:
> upvoted! @Jaex @McoreD @BrycensRanch @SemyonL @GriffinCodes

Hey @chrisatx, thanks for the upvote and for being eager to see this move forward!

Just a quick heads-up: while we’re excited to get this resolved, pinging the entire thread actually slows things down. Each notification pulls a developer out of 'the zone' or away from the code they’re currently writing to fix issues like this one.

Rest assured, the right people are watching the repo. We'll post updates here as soon as there's progress. :heart:

ahossny:
@McoreD 
I really appreciate the effort you, Jaex an all the community who are putting time into this. ShareX is software I personally rely on daily, so it’s great to see ARM64 being actively addressed now that proper hardware is available.

I completely understand the need to wait for real devices before validating fixes across architectures. For what it’s worth, I’ve been using an ARM64 laptop since they first became available, and I am eager to help contribute in my free time if that would be useful. From my perspective, the issues outlined here are all fixable with some refactoring and alignment.

Thank you again, and let me know if I can be of any assitance

Issue #8373, opened 2026-01-31
[Feature Request] Immich as a destination.


micmou:
Is there any chance we could get Immich as a destination for uploading to? I know not much of a feature request but many people who do self hosting use Immich and being able to upload directly to self hosted resources would be nice.

Gloval:
Would save a step or 3 when truying to leverage a self hosted image.

micmou:
Thanks for putting a tag on it @Jaex <3

mapperize:
@micmou There is a binary here under x64 releases if you wanna try it https://github.com/mapperize/ShareX/actions/runs/21789334479/job/62865858359

and instructions under readme here https://github.com/mapperize/ShareX/tree/develop

micmou:
Ill give it a try and let you know <3

mapperize:
> Ill give it a try and let you know <3

i realized this only works if you use immich public proxy, im working on a different version that'll hopefully be done within the next couple of days

picpo:
> [@micmou](https://github.com/micmou) There is a binary here under x64 releases if you wanna try it https://github.com/mapperize/ShareX/actions/runs/21789334479/job/62865858359
> 
> and instructions under readme here https://github.com/mapperize/ShareX/tree/develop

Hoping for a PR soon and to be merged into the main branch. I really need this feature.

mapperize:
> > [@micmou](https://github.com/micmou) There is a binary here under x64 releases if you wanna try it https://github.com/mapperize/ShareX/actions/runs/21789334479/job/62865858359
> > and instructions under readme here https://github.com/mapperize/ShareX/tree/develop
> 
> Hoping for a PR soon and to be merged into the main branch. I really need this feature.

I highkey forgot about this after my pc running immich went down from a dead PSU. Idk if I myself would finish anytime soon

BlueBull010:
I would also like to throw my vote in the bowl for this feature. As a workaround, I can of course just have ShareX drop the screenshot in a network folder which Immich is watching but a direct upload would be very nice indeed. It's just a nice-to-have of course, not critical

PR #8385, opened 2026-02-07
Add Immich Uploading Support

mapperize:
This addresses https://github.com/ShareX/ShareX/issues/8373

however I'm not sure if Jaex will like it because I use the newer HttpClient instead of the functions using the legacy HttpWebRequest. I hope at least you'll use it as a reference.


mapperize:
submitting a better pr

